Issue #24, opened 2016-01-04
Re-run pending tasks


geofffranks:
Currently Shield does not do anything with tasks that never come out of the pending state. To address this:

Add a function to ReschedulePendingTasks:
- Find all tasks in the database listed as pending.
- Cross-reference those tasks with the tasks in the scheduling queue of the supervisor. If there is a database-found task which is not in the scheduling queue, add it to the queue

Call this function at startup

filefrog:
This requires changes to the schema (hello, v3!) to track more important information about Task objects (like, what they are supposed to be doing).

PR #44, opened 2016-01-15
Fail Running Tasks on Startup

### LINKED ISSUE #30: Fail 'running' tasks on startup
When shield starts up, check for any records in the tasks table that are of state 'running'. For each found:
- Fail the task
- If the task is a 'backup' operation, and has an associated archive/restore key, trigger a purge for that archive.

jhunt:
... and purge their archives.

Fixes #30

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -81,6 +81,7 @@ func (db *DB) Exec(sql_or_name string, args ...interface{}) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


filefrog:
The purge tasks are scheduled as pending.  Once #24 is done, those purge tasks should get kicked off.

Issue #56, opened 2016-01-21
Better shield CLI help


xiujiao:
1) If the user did not target shield, can we out put something like set your target" shield -H ${shield_ip:port} [options] (command} instead of simply saying `Get https://shield/v1/targets: dial tcp: lookup shield: no such host`. 

2) If the user did set the target, but it is a wrong ip, then we will out put something like "check your shield target ip exist".  

3) when we type `shield", the help list should include set SHIELD_TARGET part.

4) Rename SHIELD_TARGET to SHIELD_API (avoid confusion which target plugin and shield_target_endpoint)

5) Formatting shield output to be more human friendly

xiujiao:
I am working on this based off of @qanx #60 PR branch

jhunt:
I just merged #60, so you should be back to working off of master.

PR #60, opened 2016-01-23
Checking for SHIELD_TARGET env var

qanx:
Output to terminal if the user hasn't specified either SHIELD_TARGET or used `shield -H`.


jhunt:
Can you resolve the conflicts on this?

I feel like it's related to #56 (which @xiujiao was working on) but I can't be sure since Github won't show the changes.

PR #61, opened 2016-01-25
Better Shield CLI Help / Environment Variable Naming

xiujiao:
Fixed resolving #56

### DIFF
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -18,7 +18,7 @@ import (
 
 func require(good bool, msg string) {
 	if !good {
-		fmt.Fprintf(os.Stderr, "\x1b[1mUSAGE:\x1b[0m\n\t%s ...\n", msg)
+		fmt.Fprintf(os.Stderr, "USAGE: %s ...\n", msg)
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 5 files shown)


cmd/shield/main.go
```diff
@@ -160,7 +167,7 @@ func main() {
 	*/
 
 	c.HelpBreak()
-	c.Dispatch("list targets", "List available backup targets",
+	c.Dispatch("list targets", "List available backup targets\r\x1b[1A\x1b[1mTARGETS:\n\x1b[0m",
```
jhunt: This breaks the semantics of how help works.  Instead of abusing the implementation of help, we should add a new method to `Command` to do something like this:

```
c.HelpGroup("TARGETS")
c.Dispatch("list targets", ... etc ...)
```

cmd/shield/main.go
```diff
@@ -127,7 +134,7 @@ func main() {
 	    ######     ##    ##     ##    ##     #######   ######
 	*/
 
-	c.Dispatch("status", "Query the SHIELD backup server for its status and version info",
+	c.Dispatch("status", "Query the SHIELD backup server for its status and version info\n",
```
jhunt: This and other newlines seem to have been added to force blank lines between groups of things?  If so, we should modify HelpBreak() to output an additional newline, rather than relying on the string-pasting implementation currently at play.

cmd/shield/main.go
```diff
@@ -523,7 +530,7 @@ func main() {
 	*/
 
 	c.HelpBreak()
-	c.Dispatch("list retention policies", "List available retention policies",
+	c.Dispatch("list retention policies", "List available retention policies\r\x1b[1A\x1b[1mPOLICIES:\n\x1b[0m",
```
jhunt: ANSI escape sequences are cute and all, but need to take into account whether or not the the current output source is in fact a terminal capable of handling ANSI CSIs.  We used [github.com/jhunt/ansi](https://github.com/jhunt/ansi) for the colorizing of confirmation and error messages in `tui`, so I'd like to see use that here as well.  I don't believe it supports bold _yet_, but that can be fixed.

cmd/shield/main.go
```diff
@@ -108,12 +112,15 @@ func main() {
...
-			fmt.Fprintf(os.Stderr, "Recommended:\n")
-			fmt.Fprintf(os.Stderr, "  Use SHIELD_TARGET environmental variable to store shield's IP:port.\n\n")
-			fmt.Fprintf(os.Stderr, "Commands:\n")
+			fmt.Fprintf(os.Stderr, "\n\x1b[1mNAME:\x1b[0m\n  shield\t\tCLI for interacting with the Shield API.\n")
+			fmt.Fprintf(os.Stderr, "\n\x1b[1mUSAGE:\x1b[0m\n  shield [options] <command>\n")
+			fmt.Fprintf(os.Stderr, "\n\x1b[1mENVIRONMENT VARIABLES:\x1b[0m\n")
+			fmt.Fprintf(os.Stderr, "  SHIELD_API\t\tused to specify the shield API's IP:port.\n")
+			fmt.Fprintf(os.Stderr, "  SHIELD_TRACE\t\tset to 'true' for trace output.\n")
+			fmt.Fprintf(os.Stderr, "  SHIELD_DEBUG\t\tset to 'true' for debug output.\n")
+			fmt.Fprintf(os.Stderr, "\n\x1b[1mINFO:\x1b[0m\n")
```
jhunt: Don't use raw ANSI escape codes unilaterally.  (see later comment for more details)

xiujiao:
ok so you are going to add that new function and change this to use it ?

cmd/shield/main.go
xiujiao: yes, that is what I did. It is better we modified HelpBreak().

PR #83, opened 2016-02-02
Redis load

geofffranks:


geofffranks:
In addition to the unit tests, verified that the redis plugin still functions properly (backup a deployment with services - both shared + dedicated, verify apps working, delete those services + their data, verify apps fail, run restore to bosh shared + dedicated, verify apps working again)

jhunt:
LGTM

jhunt:
:shipit:

PR #89, opened 2016-02-04
Failure detection fixes

geofffranks:
See commit messages for details


geofffranks:
Tested on bosh-lite by:
1. running a task that would fail, but have parseable json output + valid store key
2. running a task that would succeed but have unparseable json output
3. running a task that would succeed but have an empty store_key in the json output
4. running a task that would succeed

Verified that:
1. task  was marked as 'failed' and archive record marked as 'invalid'
2. task was marked as 'failed' and no archive record existed
3. task was marked as 'failed' and no archive record existed
4. task was marked as 'done' and archive record marked as 'valid'

PR #90, opened 2016-02-04
Fix return codes sent from agent to worker

geofffranks:


geofffranks:
Tested on bosh-lite with a shield-pipe that succeeds, a shield-pipe that sigterms itself, and a shield-pipe that exits normally with rc=3.

PR #95, opened 2016-02-12
Added human vs machine readable

qanx:


cmd/shield/forms.go
```diff
@@ -1,9 +1,9 @@
 package main
 
-import (
+/*import (
 	"fmt"
 	"strconv"
-)
+)*/
```
jhunt: Can we remove this block?

qanx: Sorry forgot this was in there - removed.

PR #106, opened 2016-02-18
Limits number of archives listed to 20 by default

### LINKED ISSUE #81: `shield ls archives` should only show X most recent archives
Right now all archives are listed regardless of their number, so when there are a large number of archives (say for a large number of services with nightly backups) running `shield ls archives` will just list all the archives even if the total # is in the thousands. Recommend only outputting the first X with an option to `--all`. and/or adding an option to let the user specify how many to list, e.g. `shield ls archives --recent 20`.

qanx:
Similar to tasks, the number of archives listed can now be limited with "limit" and the CLI uses 20 as the default.

Fixes #81

### DIFF
--- a/supervisor/archives_api.go
+++ b/supervisor/archives_api.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"net/http"
 	"regexp"
-	"strconv"
 
 	"github.com/pborman/uuid"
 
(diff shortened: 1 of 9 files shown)


supervisor/archives_api.go
```diff
@@ -26,6 +27,14 @@ func (self ArchiveAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 			desiredStatus = []string{status}
 		}
+		limit := paramValue(req, "limit", "")
+		if limit != "" {
+			limint, err := strconv.Atoi(limit)
+			if err != nil || limint <= 0 {
+				bailWithError(w, ClientErrorf("invalid limit supplied"))
+				return
+			}
+		}
```
jhunt: We should refactor this (and the other occurrence from the other day) into a `validateLimit(string) error` function that will validate the passed strings as a non-negative value and return the `ClientErrorf(...)` call.

jhunt:
Otherwise looks great!

PR #107, opened 2016-02-18
Added Read-only replicas support for mysql plugin

soutenniza:


plugin/mysql/plugin.go
```diff
@@ -21,7 +21,8 @@
 //        "mysql_user":"username-for-mysql",
 //        "mysql_password":"password-for-above-user",
 //        "mysql_host":"hostname-or-ip-of-mysql-server",
-//        "mysql_port":"port-above-mysql-server-listens-on"
+//        "mysql_port":"port-above-mysql-server-listens-on",
+//				"mysql_read_replica":"hostname-or-ip-of-mysql-replica-server", OPTIONAL
```
geofffranks: Can you drop the trailing comma, and add a '#' in front of OPTIONAL, so that when this is copy/pasted from godoc.org, it's valid json. Also, make the whitespace match up with the above lines?

PR #116, opened 2016-02-25
Use godep to make Makefile environment independant

bodymindarts:


jhunt:
I don't believe we still need this, with vendoring

bodymindarts:
agreed

PR #117, opened 2016-02-25
Combined filter.Query and Filter.Args to avoid further change anomalies

thomasmmitchell:


jhunt:
This is perfect. :shipit:

PR #121, opened 2016-03-07
Web UI

jhunt:
This commit introduces a fully-hosted web user interface for viewing
tasks, managing jobs / schedules / retention policies / stores /
targets, restoring archives, etc.


jhunt:
I was surprised as well.  Couldn't find one that allowed 18-months rolling.

PR #123, opened 2016-03-07
shield-pipe: Run `validate` on all executed plugins

jhunt:
Before we try to run the plugins in a pipeline, shield-pipe will now
validate the endpoint configurations for each plugin involved, and bail
out if any of them are invalid.

Also, cleaned up some output so that task logs are prettier.

screen shot 2016-03-07 at 10 48 18 am

### DIFF
--- a/bin/shield-pipe
+++ b/bin/shield-pipe
@@ -35,52 +35,107 @@
 #     https://code.google.com/p/go/issues/detail?id=2266
 #
 
+header() {
+	echo >&2 $*
+	echo $* | sed -e 's/./=/g' >&2
+}
+
+say() {
+	echo >&2 $*
+}
+
+ok() {
+	echo >&2 "OK"
+	echo >&2
+}
+
+fail() {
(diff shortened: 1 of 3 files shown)


bin/shield-pipe
```diff
@@ -35,52 +35,104 @@
...
+	header "Running restore task (using bzip2 compression)"
 	set -o pipefail
 	${SHIELD_STORE_PLUGIN} retrieve -k "${SHIELD_RESTORE_KEY}" -e "${SHIELD_STORE_ENDPOINT}" | bunzip2 | \
 		${SHIELD_TARGET_PLUGIN} restore -e "${SHIELD_TARGET_ENDPOINT}"
-	exit $?
+	exiting $?
 	;;
 
 (purge)
 	needenv SHIELD_RESTORE_KEY
```
geofffranks: need more envs here

jhunt: Done and pushed.

PR #133, opened 2016-03-15
Postgres Plugin Option of Specific Database

dstarner15:
**Changes**
- In configuration, a field `pg_database` was added as an option. If included, it will backup and run a `pg_dump` on the database specified in the `pg_database` key. If excluded or left as `""`, then Shield will run a `pg_dumpall` as before.

**Internally Changed**
- Postgres Plugin for a Target

### DIFF
--- a/plugin/postgres/plugin.go
+++ b/plugin/postgres/plugin.go
@@ -22,6 +22,7 @@
 //        "pg_password":"password-for-above-user",
 //        "pg_host":"hostname-or-ip-of-pg-server",
 //        "pg_port":"port-above-pg-server-listens-on"
+//		  "pg_database": "name-of-db-to-backup"
 //    }
 //
 // BACKUP DETAILS
@@ -95,6 +96,7 @@ type PostgresConnectionInfo struct {
 	User     string
 	Password string
 	Bin      string
+	Database string
 }
 
 func (p PostgresPlugin) Meta() PluginInfo {
@@ -154,8 +156,20 @@ func (p PostgresPlugin) Backup(endpoint ShieldEndpoint) error {
 	}
 
 	setupEnvironmentVariables(pg)
-	cmd := fmt.Sprintf("%s/pg_dumpall -c --no-password", pg.Bin)
-	DEBUG("Executing: `%s`", cmd)
+
+	// If user specified specific database to backup
+	db_name := pg.Database
+
+	// Check if db_name is valid
+	if db_name != "" || db_name != nil {
+		// Run dump all on the specified db
+		cmd := fmt.Sprintf("%s/pg_dumpall -d %s -c --no-password", pg.Bin, db_name)
+		DEBUG("Executing: `%s`", cmd)
+	} else {
+		// Else run dump on all
+		cmd := fmt.Sprintf("%s/pg_dumpall -c --no-password", pg.Bin)
+		DEBUG("Executing: `%s`", cmd)
+	}
 	return Exec(cmd, STDOUT)
 }
 
@@ -245,6 +259,9 @@ func pgConnectionInfo(endpoint ShieldEndpoint) (*PostgresConnectionInfo, error) 
 	}
 	DEBUG("PGPORT: '%s'", port)
 
+	database, err := endpoint.StringValueDefault("pg_database")
+	DEBUG("PGDATABASE: '%s'", database)
+
 	bin := "/var/vcap/packages/postgres-9.4/bin"
 	DEBUG("PGBINDIR: '%s'", bin)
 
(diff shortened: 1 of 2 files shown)


plugin/postgres/plugin.go
```diff
@@ -22,6 +22,7 @@
 //        "pg_password":"password-for-above-user",
 //        "pg_host":"hostname-or-ip-of-pg-server",
 //        "pg_port":"port-above-pg-server-listens-on"
```
geofffranks: Can you add a comma to the end of this comment, so it generates valid json, now that you've added a new field?

plugin/postgres/plugin.go
```diff
@@ -154,8 +156,21 @@ func (p PostgresPlugin) Backup(endpoint ShieldEndpoint) error {
...
-	DEBUG("Executing: `%s`", cmd)
+
+	// If user specified specific database to backup
+	db_name := pg.Database
+
+	cmd := ""
+	// Check if db_name is valid
+	if db_name != "" {
+		// Run dump all on the specified db
+		cmd = fmt.Sprintf("%s/pg_dump %s", pg.Bin, db_name)
```
geofffranks: We're going to want to keep the `-c --no-password` flags for pg_dump here as well

geofffranks:
Can you also add support for pg_database in the Validate() function?

jhunt:
@dstarner15 any updates on @geofffranks feedback above?

plugin/postgres/plugin.go
```diff
@@ -141,6 +143,13 @@ func (p PostgresPlugin) Validate(endpoint ShieldEndpoint) error {
 		ansi.Printf("@G{\u2713 pg_password}  @C{%s}\n", s)
 	}
 
+	s, err = endpoint.StringValue("pg_database")
+	if err != nil {
+		ansi.Printf("@R{\u2717 pg_database  %s}\n", err)
```
geofffranks: I *think* this should be using StringValueDefault, similar to port, so that if this is not specified, it doesn't cause an error, but does cause an if the value is not a string. @jhunt will know for sure though

jhunt: Yes, this should match the other place in the proposed changes where `StringValueDefault` is called.

plugin/postgres/plugin.go
```diff
@@ -154,8 +163,21 @@ func (p PostgresPlugin) Backup(endpoint ShieldEndpoint) error {
 	}
 
 	setupEnvironmentVariables(pg)
-	cmd := fmt.Sprintf("%s/pg_dumpall -c --no-password", pg.Bin)
-	DEBUG("Executing: `%s`", cmd)
+
+	// If user specified specific database to backup
+	db_name := pg.Database
```
jhunt: Is there a compelling reason to use a temporary variable?  I see no transformations on the `db_name` copy of `pg.Database`...

plugin/postgres/plugin.go
```diff
@@ -21,7 +21,8 @@
 //        "pg_user":"username-for-postgres",
 //        "pg_password":"password-for-above-user",
 //        "pg_host":"hostname-or-ip-of-pg-server",
-//        "pg_port":"port-above-pg-server-listens-on"
+//        "pg_port":"port-above-pg-server-listens-on",
+//        "pg_database": "name-of-db-to-backup"
```
jhunt: We should add a note that `pg_database` is optional, and explain what the semantics are when present and when omitted.

plugin/postgres/plugin.go
jhunt: Additionally, this code needs to take into account the "all databases" vs "this database" distinction, and print out a message like:

```
pg_database         none (all databases will be backed up)
```

to let the operator know that all databases will be backed up.

plugin/postgres/plugin.go
```diff
@@ -154,8 +163,21 @@ func (p PostgresPlugin) Backup(endpoint ShieldEndpoint) error {
...
+	cmd := ""
+	// Check if db_name is valid
+	if db_name != "" {
+		// Run dump all on the specified db
+		cmd = fmt.Sprintf("%s/pg_dump %s -c --no-password", pg.Bin, db_name)
+		DEBUG("Executing: `%s`", cmd)
+	} else {
+		// Else run dump on all
+		cmd = fmt.Sprintf("%s/pg_dumpall -c --no-password", pg.Bin)
+		DEBUG("Executing: `%s`", cmd)
```
jhunt: These two DEBUG statements can go outside of the conditional blocks, since they are identical in both cases.

PR #138, opened 2016-04-20
Added Initial Authentication Support.

geofffranks:
Authentication is now required for all requests. Requests are
checked for authentication in the following manner:

1. If the `X-SHIELD-TOKEN` header is set, and matches a token
   listed in the `auth.api_tokens` map, the request is allowed.
2. If the `auth.oauth.provider` property is set, enables OAuth2
   authentication, and disables Basic authentication. Corresponding
   `auth.oauth` properties must also be set (session db info, secrets,
   etc).
   i.  If the user has a valid session that has previously authenticated,
       let them in.
   ii. If not, redirect them to the OAuth provider, allow them to login
       and authorize the SHIELD app, with a callback to SHIELD to verify
       authentication, and lastly redirect to their originally requested page.
3. If `auth.oauth.provider` was set, check the request for HTTP Basic
   authentication credentials, based off the `auth.basic.user` and
   `auth.basic.password` properties. If those properties weren't set,
   they were given a default value.
4. If none of the above worked for protected URIs, return a 401 to the user.

The `shield` CLI has been updated to coincide with this:
1. `shield` Will detect authentication requests, and prompt users for credentials/bearer tokens
2. `shield` Will save authorization tokens into the user's `.shield_config` file.
3. `SHIELD_API` and the `-H` flag for `shield` are now deprecated. If specified, they
   will auto-vivify a `default` backend for shield, in the `.shield_config`.
4. `shield backends`, `shield use backend`, `shield create backend` are used to manage
   connecting to multiple SHIELD backend servers.

The SHIELD WebUI has been updated as well:
1. Normal authentication works seamlessly, users are prompted for credentials
   when basic authentication is enabled, or are redirected to the OAuth2 provider,
   when OAuth2 is enabled.
2. If an API request from the WebUI triggers an authentication event, the currently
   viewed page is reloaded, to work-around failures of AJAX calls trying to authenticate
   via OAuth2.


api/config.go
```diff
@@ -0,0 +1,121 @@
...
+func (cfg *Config) Save() error {
+	data, err := yaml.Marshal(cfg)
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(cfg.Path, data, 0644)
+	if err != nil {
+		return err
+	}
```
jhunt: Change this to 0600, so that auth credentials are not visible to other users, or other users in the same group (for Jumpboxen without sudo).

Also, is there any way to do this as a 2-stage creat+rename?  That way we always creat the file with 0600, and we never put creds on-disk with open permissions (in case the user was silly and chmod'd the file)

api/config_test.go
```diff
@@ -0,0 +1,296 @@
...
+aliases: {}
+`
+
+			Expect(cfg.Save()).Should(Succeed())
+
+			data, err := ioutil.ReadFile(tempFile.Name())
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(string(data)).Should(Equal(expectedCfg))
+
+			Expect(os.Remove(tempFile.Name())).Should(Succeed())
```
jhunt: Do we really want the tests to fail if there is a problem with removing the tempfile used for testing?  It's not really a reflection of the code under test...

api/config_test.go
```diff
@@ -0,0 +1,296 @@
...
+					"shield1": "http://localhost",
+					"shield2": "http://localhost:8080",
+					"invalid": "http://google.com",
+				},
+			}
+		})
+		It("Returns empty string if no current backend set", func() {
+			Expect(cfg.BackendURI()).Should(Equal(""))
+		})
+		It("Returns an empty string if current backend is an alias is set to an invalid backend", func() {
```
jhunt: Verbiage problem? "backend is an alias is set to"?

api/url.go
```diff
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/howeyc/gopass"
```
jhunt: Can you take a look at https://godoc.org/golang.org/x/crypto/ssh/terminal and see what you think?  I used it on safe and it worked okay.  If this works better, I'd like to use it in safe. See https://github.com/jhunt/safe/blob/master/prompt/prompt.go

cmd/shield/main.go
```diff
@@ -56,6 +58,10 @@ func main() {
 		Before:    getopt.StringLong("before", 'B', "", "Only show archives that were taken before the given date, in YYYYMMDD format."),
 		To:        getopt.StringLong("to", 0, "", "Restore the archive in question to a different target, specified by UUID"),
 		Limit:     getopt.StringLong("limit", 0, "", "Display only the X most recent tasks or archives"),
+
+		Config:   getopt.StringLong("config", 'c', os.Getenv("HOME")+"/.shield_config", "Specify the shield config file to use"),
```
jhunt: Can we be a bit more explicit in the help, and call it the `~/.shield_config`

cmd/shield/main.go
```diff
@@ -159,6 +151,78 @@ func main() {
...
+			if len(args) != 2 {
+				return fmt.Errorf("Invalid 'create backend' syntax: `shield backend <name> <uri>")
+			}
+			err := Cfg.AddBackend(args[1], args[0])
+			if err != nil {
+				return err
+			}
+			Cfg.Save()
+
+			ansi.Fprintf(os.Stdout, "Successfully created backend '@G{%s}', pointing to '@G{%s}'\n\n", args[0], args[1])
```
jhunt: Can we implicitly use the backend after we create it?  To avoid the following churn:

```
shield create backend prod http://prod.shield.us
shield use backend prod
```

cmd/shield/main.go
```diff
@@ -1465,6 +1529,50 @@ func main() {
...
+			ansi.Fprintf(os.Stderr, "@Y{WARNING: -H, --host, and the SHIELD_API environment variable have been deprecated and will be removed in a later release.} Use `shield backend` instead\n")
+		}
+
+		if len(Cfg.Backends) == 0 {
+			backend := os.Getenv("SHIELD_API")
+			if *options.Shield != "" {
+				backend = *options.Shield
+			}
+			ansi.Fprintf(os.Stderr, "@C{Initializing `default` backend as `%s`}\n", backend)
+			err := Cfg.AddBackend(backend, "default")
```
jhunt: backend could be empty here, if the caller didn't supply `-h`, and there is no `$SHIELD_API` environment variable.

We should probably also test what happens when you `AddBackend("", "alias")`, in the Config tests.

cmd/shield/main.go
```diff
@@ -1465,6 +1529,50 @@ func main() {
...
+			}
+			Cfg.UseBackend("default")
+		}
+
+		if Cfg.BackendURI() == "" {
+			ansi.Fprintf(os.Stderr, "@R{No backend targeted. Use `shield list backends` and `shield backend` to target one}\n")
+			os.Exit(1)
+		}
+
+		if *options.User != "" || *options.Password != "" {
```
jhunt: Can we drop these options altogether? I don't like having the `--password` as an argument, since it will show up in `~/.bash_history` and in `ps` output.  If we want to be able to _reset_ credentials in Basic Auth mode, we should either prompt on 403 (not preferred) or support a `shield auth` command that prompts for user / password

supervisor/auth.go
```diff
@@ -0,0 +1,69 @@
...
+func (ba BasicAuthenticator) RequireAuth(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("WWW-Authenticate", "Basic realm=\"shield\"")
+	w.WriteHeader(401)
+	w.Write([]byte("Unauthorized"))
+}
+
+func Authenticate(tokens map[string]string, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		for name, token := range tokens {
+			log.Debugf("Checking Tokens")
```
jhunt: minor: maybe add some per-token debugging here instead of just "Checking Tokens" over and over again

supervisor/jwt.go
```diff
@@ -0,0 +1,38 @@
...
+		w.Write([]byte("Unable to generate authentication token"))
+		return
+	}
+	w.Write([]byte("Bearer " + token))
+}
+
+func (jc JWTCreator) GenToken(user interface{}, maxAge int) (string, error) {
+	token := jwt.New(jwt.SigningMethodRS256)
+	token.Claims["expiration"] = maxAge
+	// FIXME store org info for authorization
```
jhunt: FIXME?

supervisor/oauth.go
```diff
@@ -0,0 +1,128 @@
...
+
+	user, err := gothic.CompleteUserAuth(w, r)
+	if err != nil {
+		w.WriteHeader(403)
+		w.Write([]byte("UnOAuthorized"))
+		return
+	} else {
+		log.Infof("Successful login for %s", user)
+	}
+	//FIXME: do some sort of authorization on the user to restrict access to more than any user with a password
```
jhunt: FIXME?

webui/js/app.js
```diff
@@ -11,7 +11,11 @@
       AJAX = $.ajax(params)
         .always(function() { AJAX = null })
         .fail(function(jqxhr) {
-          notify('danger', "Backend request failed: "+jqxhr.status.toString()+' '+jqxhr.statusText);
+          // reload the browser if we had an ajax call needing oauth, as cors will prevent this + cause errors
+          if (jqxhr.status == 401 && jqxhr.getResponseHeader("WWW-Authenticate").toLowerCase().StartsWith == "bearer") {
+            location.reload(true)
+          }
+          notify('danger', "Backend request failed: "+jxhr.status.toString()+' '+jqxhr.statusText);
```
jhunt: This should be `jqxhr`, not `jxhr`...

jhunt:
Do we want to keep the CLI help documentation about the `SHIELD_API` environment variables and the `-H` and `--host` arguments, or should we at least put a note about deprecation?

Overall, looks great!

PR #139, opened 2016-04-25
Vendoring

geofffranks:
Converted to golangs built-in vendoring mechanism. 

godep still works as normal, but manages data in the vendor directory instead of Godeps/_workspace.

Also updated the Makefile to support saving/restoring dependencies and handling the new directory structures without too much pain.


jhunt:
Remove the '@' on the Makefile recipes and this can be merged.

PR #142, opened 2016-04-26
Change the testdev script so it also kicks off the shield-agent

jrbudnack:
Just putting this up here so we can chat about it tomorrow.

A couple of discussion points before you merge this:
- I had to do some handling of the temp directory such that both the agent and the daemon could share it.  This is mainly so that they can share the generated keypair.
- The script seems to always assume that a "make" has been performed, building the binaries so they can be run.  Should we continue this assumption?  I just ask because I'm also using the current directory as the plugin directory, as that's where make seems to deposit them.
- Another assumption I make is that the appropriate Postgres tools are accessible from the expected /var/vcap/packages directory (since that's the path hard-coded into the plugin).  On my own laptop, I simply created a dummy /var/vcap/packages hierarchy and symbolically linked to the required Postgres utilities.  I can script this as well, but that in and of itself makes other assumptions.

Anyway, I tested it out and was able to run a backup, pointing at the shield-agent that this spins up.  Let me know if you have any questions.

Jeremy


jhunt:
I'm okay with the manual `/var/vcap/packages/...` work, since I view the testdev script more as enabling SHIELD development itself (core / agent / CLI / WebUI), and less about plugin development.  The `fs` plugin works great as a low-dependency plugin for testing the rest of the system.

jhunt:
As for the shared keys thing, the agent shouldn't need to share keys with the core.  The only thing the agent needs from the core is the public key, which is available via the API.  See https://github.com/starkandwayne/shield-boshrelease/blob/master/jobs/shield-agent/templates/bin/ctl#L25-L29

If you want to (totally optional) you can make use of that, which nicely exercises the meta API as part of spinning up testdev, and more accurately portrays how we do auto-provisioning of agents in the wild.

jhunt:
Looks like @geofffranks introduced conflicting code in his PRs.  Can you merge/resolve and re-submit?

jrbudnack:
@jhunt - Merged

PR #148, opened 2016-05-05
Enable UAA authentication/authorization support

geofffranks:
Added support for the UAA to be used as an OAauth2 provider
for SHIELD, to handle authentication/authorization. If used,
you must specify the `auth.oauth.provider_url` configuration property,
to point at the UAA being used.

Authorization is performed by seeing if the authenticated user is
in a member of a UAA group(s) listed in `auth.oauth.authorization.orgs`.

### DIFF
--- a/supervisor/oauth_verifiers.go
+++ b/supervisor/oauth_verifiers.go
@@ -1,16 +1,20 @@
 package supervisor
 
 import (
+	"fmt"
+	"github.com/geofffranks/botta"
 	"github.com/google/go-github/github"
 	"github.com/markbates/goth"
 	"github.com/starkandwayne/goutils/log"
 	"net/http"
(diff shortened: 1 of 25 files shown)


supervisor/oauth_verifiers.go
```diff
@@ -45,18 +41,156 @@ func (gv *GithubVerifier) Verify(user goth.User, c *http.Client) bool {
...
+			}
+			if uid == user.UserID {
+				membership = append(membership, name)
+			}
+		}
+	}
+	return map[string]interface{}{"Groups": membership}, nil
+}
+
+// FIXME: tests.. ghalrdfsa
```
jhunt: I assume this FIXME can be removed?

PR #167, opened 2016-05-25
Scality Isn't Quite S3

thomasmmitchell:
You may have thought it was S3, but it wasn't, and you were wrong.

There is now a plugin specifically for Scality. It uses a fork of the minio-go library that has the added feature of being able to set a max part size for multipart object "put"s. The Scality plugin always uses the V2 auth signature, and sets the max part size to 75MB, which is confirmed to work with Scality.

Otherwise, its pretty much the S3 plugin without the unnecessary parts. I would know - I copied that file.

### DIFF
--- a/plugin/scality/plugin.go
+++ b/plugin/scality/plugin.go
@@ -1,0 +1,335 @@
+// The `scality` plugin for SHIELD is intended to be a back-end storage
+// plugin, wrapping the S3 API that Scality exposes, slightly altered to work
+// within the confines of the ways that Scality differs from S3. It uses the
+// version 2 signature version, and places a lower limit on the part size
+// of a multipart object `put` operation.
+//
+// PLUGIN FEATURES
+//
+// This plugin implements functionality suitable for use with the following
+// SHIELD Job components:
+//
+//  Target: no
+//  Store:  yes
+//
+// PLUGIN CONFIGURATION
+//
+// The endpoint configuration passed to this plugin is used to determine
+// how to connect to Scality, and where to place/retrieve the data once connected.
+// your endpoint JSON should look something like this:
+//
+//    {
+//        "scality_host":        "your-scality-host",
+//        "access_key_id":       "your-access-key-id",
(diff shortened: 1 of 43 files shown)


thomasmmitchell:
Clearly a vendoring issue - looking into it

plugin/scality/plugin.go
```diff
@@ -0,0 +1,335 @@
...
+import (
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/jhunt/ansi"
+	minio "github.com/starkandwayne/minio-go"
```
geofffranks: Maybe add a note that we forked minio-go, and why we need our forked version?

plugin/scality/plugin.go
```diff
@@ -0,0 +1,335 @@
...
+	client, err := scal.Connect()
+	if err != nil {
+		return "", err
+	}
+
+	path := scal.genBackupPath()
+	plugin.DEBUG("Storing data in %s", path)
+
+	minio.SetMaxPartSize(1024 * 1024 * 75)
+	// FIXME: should we do something with the size of the write performed?
```
geofffranks: Should this FIXME get fixed?

plugin/scality/plugin.go
thomasmmitchell: The blurb at the top of the file sort of explains that

plugin/scality/plugin.go
thomasmmitchell: Seems incorrectly marked as a FIXME. Nothing is broken. More of a... TODO? Either way, maybe print it to output? What do you think?

thomasmmitchell: Whatever we decide, I'll mirror it in the S3 plugin

plugin/scality/plugin.go
thomasmmitchell: I'll make a specific note at the import line.

PR #177, opened 2016-06-09
Fixes Postgres Restore Bug

thomasmmitchell:
- Fixes  a bug where a Postgres dump file with a line longer than 64K could not be read by the scanner in the Postgres restore plugin.
- The plugin now returns non-0 codes on errors related to reading the file in.
- Fixed an off-by-one error in the debug logs when it says how many lines were in the dump file that was restored.

### DIFF
--- a/plugin/postgres/plugin.go
+++ b/plugin/postgres/plugin.go
@@ -199,28 +199,54 @@ func (p PostgresPlugin) Restore(endpoint ShieldEndpoint) error {
 	if err != nil {
 		return err
 	}
-	go func(out io.WriteCloser, in io.Reader) {
+	scanErr := make(chan error)
+	cmdErr := make(chan error)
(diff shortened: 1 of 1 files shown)


plugin/postgres/plugin.go
```diff
@@ -271,7 +295,8 @@ func pgConnectionInfo(endpoint ShieldEndpoint) (*PostgresConnectionInfo, error)
 	database, err := endpoint.StringValueDefault("pg_database", "")
 	DEBUG("PGDATABASE: '%s'", database)
 
-	bin := "/var/vcap/packages/postgres-9.4/bin"
+	//bin := "/var/vcap/packages/postgres-9.4/bin"
+	bin := "/var/vcap/packages/postgresql_9.3/bin"
```
geofffranks: Was this intended?

thomasmmitchell:
Carry on

PR #181, opened 2016-06-15
refactor to get back a task ID when run adhoc backup job

karampok:
Hi!
As we discussed shortly in slack, i would like to be able to get back an taskid after running adhoc a backup job. I spent some time to fix this behavior. Tried to keep the changes to minimum.
What do you think? 

Thx

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -94,6 +94,14 @@ func UnpauseJob(id uuid.UUID) error {
 	return ShieldURI("/v1/job/%s/unpause", id).Post(nil, "{}")
 }
 
-func RunJob(id uuid.UUID, ownerJSON string) error {
-	return ShieldURI("/v1/job/%s/run", id).Post(nil, ownerJSON)
+//If the string returned is the empty string but the error returned is nil, then
+//it is most likely that the deployed version of the backend does not support
+//handing back the uuid for an adhoc task.
+func RunJob(id uuid.UUID, ownerJSON string) (string, error) {
+	respMap := make(map[string]string)
+	err := ShieldURI("/v1/job/%s/run", id).Post(&respMap, ownerJSON)
+	if err != nil {
+		return "", err
+	}
+	return respMap["task_uuid"], nil
 }
--- a/db/tasks.go
+++ b/db/tasks.go
@@ -24,25 +24,31 @@ const (
 )
 
 type Task struct {
-	UUID           uuid.UUID `json:"uuid"`
-	Owner          string    `json:"owner"`
-	Op             string    `json:"type"`
-	JobUUID        uuid.UUID `json:"job_uuid"`
-	ArchiveUUID    uuid.UUID `json:"archive_uuid"`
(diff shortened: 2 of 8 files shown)


karampok:
The doc has not been updated. Let me know first if the changes can be integrated, and will change  also that.

api/jobs.go
```diff
@@ -1,8 +1,6 @@
 package api
 
-import (
-	"github.com/pborman/uuid"
-)
+import "github.com/pborman/uuid"
```
jhunt: Did you do this manually, or did `go fmt` try to be "helpful"?

db/tasks.go
```diff
@@ -176,8 +176,7 @@ func (db *DB) GetTask(id uuid.UUID) (*Task, error) {
 	return r[0], nil
 }
 
-func (db *DB) CreateBackupTask(owner string, job *Job) (*Task, error) {
-	id := uuid.NewRandom()
+func (db *DB) CreateBackupTask(owner string, job *Job, id uuid.UUID) (*Task, error) {
```
jhunt: The `*Task* that this function previously returned should have had the generated UUID in it, so I don't think you need to forcibly impose a UUID from outside on the creation.  The proposed change also makes it so that every caller of `CreateBackupTask()` needs to be cognizant not to reuse UUIDs accidentally, which makes the overall system a little more brittle.

jhunt: see https://github.com/starkandwayne/shield/blob/master/db/tasks.go#L27

api/jobs.go
karampok: the vim go plugin should run fmt automatically. I will double check

thomasmmitchell: I can verify that the vim plugin does this formatting when you save the file.

db/tasks.go
karampok: If you are not "forcibly impose a UUID" then it is not possible to return back to the caller a ID.
This is running on the jobapi part, which sends  the task over a channel which supervisor. The task is is created on the other side of the channel. Do you think another solution to have available the taskid?

Agree that every caller to createBackuTask has to do uuid.NewUuid() which is not optimal.

karampok: Maybe we could approach "if uuid is not defined, then generate a new one" ?
Alternative we could split the function into
CreateBackupTaskWithID(...., id )
CreateBackupTas()  as it is

api/jobs.go
jhunt: Odd.  `go fmt` did not do this for me when I tested it.

I just want to avoid commit churn as different devs environments fight over "the one true import style" (something that `go fmt` was _supposed_ to solve...)

karampok:
I modified the changes so to introduce less changes in the normal (non-adhoc) case.

jhunt:
After some intense work in #186 I think we can do what this PR was hoping to accomplish.

karampok:
Yes this can be closed, https://github.com/starkandwayne/shield/pull/186 covers it in a better way.

PR #183, opened 2016-06-16
script for setting up testdev with things and stuff

### LINKED ISSUE #182: Add some basic data to testdev script
It would be nice if new local dev environments came pre-equipped with a schedule, retention policy, local fs store / target and a job to tie them all together.

thomasmmitchell:
Fixes #182 

Right now its a standalone script. If you ALWAYS want it in testdev, I can call it from the bottom of the testdev script.

### DIFF
--- a/bin/dev-setup
+++ b/bin/dev-setup
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+set -e 
+
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+
+#if no cli, make cli
+if [[ ! -f "../shield" ]]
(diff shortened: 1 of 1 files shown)


thomasmmitchell:
Tomorrow morning I'll switch the store and target over to local fs if its really important. I really just took the stuff out of what I had put into the help dialogues.

jhunt:
Can you investigate using bin/bkp instead of the SHIELD CLI (no build required) and also add a store/target using the fs plugins and the tempdir that bin/testdev sets up?

PR #186, opened 2016-06-21
Adhoc task uuid

thomasmmitchell:
The daemon now returns a task uuid for adhoc tasks (`run job` and `restore archive`).
The CLI will display these if the backend supports it.

### DIFF
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -1498,6 +1498,9 @@ func main() {
 				})
 			} else {
 				OK("Scheduled immediate run of job")
+				if taskUUID != "" {
+					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
+				}
 			}
 
 			return nil
@@ -1843,6 +1846,9 @@ func main() {
 				})
 			} else {
 				OK("Scheduled immediate restore of archive '%s' %s", id, targetMsg)
+				if taskUUID != "" {
+					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
+				}
 			}
 
 			return nil
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -1486,8 +1486,8 @@ func main() {
 				return err
 			}
 
-			var taskUUID string = ""
-			if taskUUID, err = RunJob(id, string(b)); err != nil {
+			taskUUID, err := RunJob(id, string(b))
+			if err != nil {
 				return err
 			}
 
@@ -1497,6 +1497,7 @@ func main() {
 					"task_uuid": taskUUID,
 				})
 			} else {
+				//`OK` handles raw check
 				OK("Scheduled immediate run of job")
 				if taskUUID != "" {
 					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
@@ -1830,8 +1831,8 @@ func main() {
 				return err
 			}
 
-			var taskUUID string = ""
-			if taskUUID, err = RestoreArchive(id, string(b)); err != nil {
+			taskUUID, err := RestoreArchive(id, string(b))
+			if err != nil {
 				return err
 			}
 
@@ -1845,6 +1846,7 @@ func main() {
 					"task_uuid": taskUUID,
 				})
 			} else {
+				//`OK` handles raw checking
 				OK("Scheduled immediate restore of archive '%s' %s", id, targetMsg)
 				if taskUUID != "" {
 					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -1459,6 +1459,8 @@ func main() {
 	c.Dispatch("run job", "Schedule an immediate run of a backup job",
 		func(opts Options, args []string, help bool) error {
 			if help {
+
+				MessageHelp("Note: If raw mode is specified and the targeted SHIELD backend does not support handing back the task uuid, the task_uuid in the JSON will be the empty string")
 				FlagHelp(`A string partially matching the name of a job to run 
 				or a UUID exactly matching the UUID of a job to run.
 				Not setting this value explicitly will default it to the empty string.`,
@@ -1491,13 +1493,12 @@ func main() {
 				return err
 			}
 
-			if taskUUID != "" && *opts.Raw {
+			if *opts.Raw {
 				RawJSON(map[string]interface{}{
 					"ok":        "Scheduled immediate run of job",
 					"task_uuid": taskUUID,
 				})
 			} else {
-				//`OK` handles raw check
 				OK("Scheduled immediate run of job")
 				if taskUUID != "" {
 					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
@@ -1787,6 +1788,7 @@ func main() {
 	c.Dispatch("restore archive", "Restore a backup archive",
 		func(opts Options, args []string, help bool) error {
 			if help {
+				MessageHelp("Note: If raw mode is specified and the targeted SHIELD backend does not support handing back the task uuid, the task_uuid in the JSON will be the empty string")
 				FlagHelp(`Outputs the result as a JSON object.`, true, "--raw")
 				FlagHelp(`A UUID assigned to a single archive instance`, false, "<uuid>")
 				HelpKMacro()
@@ -1840,7 +1842,7 @@ func main() {
 			if params.Target != "" {
 				targetMsg = fmt.Sprintf("to target '%s'", params.Target)
 			}
-			if taskUUID != "" && *opts.Raw {
+			if *opts.Raw {
 				RawJSON(map[string]interface{}{
 					"ok":        fmt.Sprintf("Scheduled immediate restore of archive '%s' %s", id, targetMsg),
 					"task_uuid": taskUUID,
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -1463,6 +1463,7 @@ func main() {
 				or a UUID exactly matching the UUID of a job to run.
 				Not setting this value explicitly will default it to the empty string.`,
 					false, "<job>")
+				JSONHelp(`{"ok":"Scheduled immediate run of job","task_uuid":"143e5494-63c4-4e05-9051-8b3015eae061"}`)
 				HelpKMacro()
 				return nil
 			}
@@ -1485,11 +1486,20 @@ func main() {
 				return err
 			}
 
-			if err := RunJob(id, string(b)); err != nil {
+			var taskUUID string = ""
+			if taskUUID, err = RunJob(id, string(b)); err != nil {
 				return err
 			}
 
-			OK("Scheduled immediate run of job")
+			if taskUUID != "" && *opts.Raw {
+				RawJSON(map[string]interface{}{
+					"ok":        "Scheduled immediate run of job",
+					"task_uuid": taskUUID,
+				})
+			} else {
+				OK("Scheduled immediate run of job")
+			}
+
 			return nil
 		})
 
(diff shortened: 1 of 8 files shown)


thomasmmitchell:
need to alter tests to match new behavior.

also, supervisor tests appear to hang indefinitely on failure. probably an eventually statement that will never close.

cmd/shield/main.go
```diff
@@ -1485,11 +1486,23 @@ func main() {
 				return err
 			}
 
-			if err := RunJob(id, string(b)); err != nil {
+			var taskUUID string = ""
+			if taskUUID, err = RunJob(id, string(b)); err != nil {
```
jhunt: minor nit, but here I think I would stop using the `if <initializer>; <conditional> { ... }` construct and just assign `taskUUID, err := ...`

thomasmmitchell: We're not saving a line here anymore, so yeah, you're right.

cmd/shield/main.go
```diff
@@ -1485,11 +1486,23 @@ func main() {
...
+					"ok":        "Scheduled immediate run of job",
+					"task_uuid": taskUUID,
+				})
+			} else {
+				OK("Scheduled immediate run of job")
+				if taskUUID != "" {
+					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
+				}
+			}
+
```
jhunt: Is there a case where *opts.Raw is true, but taskUUID is still an empty string (a case not caught by the error handling above?)  It kind of looks like that's a hole in this if/else block, which would lead `--raw` usage to return non-JSON.

cmd/shield/main.go
```diff
@@ -1817,15 +1830,27 @@ func main() {
 				return err
 			}
 
-			if err := RestoreArchive(id, string(b)); err != nil {
+			var taskUUID string = ""
+			if taskUUID, err = RestoreArchive(id, string(b)); err != nil {
```
jhunt: likewise here, drop the combined if-initializer-condition and just do the assignment outright.

cmd/shield/main.go
```diff
@@ -1817,15 +1830,27 @@ func main() {
...
+					"ok":        fmt.Sprintf("Scheduled immediate restore of archive '%s' %s", id, targetMsg),
+					"task_uuid": taskUUID,
+				})
+			} else {
+				OK("Scheduled immediate restore of archive '%s' %s", id, targetMsg)
+				if taskUUID != "" {
+					ansi.Printf("To view task, type @B{shield show task %s}\n", taskUUID)
+				}
+			}
+
```
jhunt: as above, it _looks_ like the if/else could lead to a case where `--raw` returns non-JSON

cmd/shield/main.go
thomasmmitchell: The function `OK` tests for the --raw flag, and will therefore print JSON in raw mode.

supervisor/archives_api.go
```diff
@@ -92,17 +92,25 @@ func (self ArchiveAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			Owner:        params.Owner,
+			TargetUUID:   tid,
+			ArchiveUUID:  id,
+			RestoreKey:   archive.StoreKey,
+			TaskUUIDChan: taskChan,
 		}
 
-		w.WriteHeader(200)
-		JSONLiteral(w, fmt.Sprintf(`{"ok":"scheduled"}`))
+		var taskUUID string = <-taskChan
```
jhunt: This should be a `:=` assignment without the `var` keyword or `string` type declaration

cmd/shield/main.go
thomasmmitchell: :+1:

supervisor/archives_api.go
```diff
@@ -92,17 +92,25 @@ func (self ArchiveAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		if taskUUID == "" {
+			w.WriteHeader(500)
+			JSONLiteral(w, fmt.Sprintf(`{"ok":"error"}`))
+		} else {
+			w.WriteHeader(200)
+			JSONLiteral(w, fmt.Sprintf(`{"ok":"scheduled","task_uuid":"%s"}`, taskUUID))
```
jhunt: Do we use `JSONLiteral` like this elsewhere?  If not, we should use real JSON marshaling.  If so, we should put a ticket in to use real JSON marshaling globally.

supervisor/jobs_api.go
```diff
@@ -142,13 +142,23 @@ func (self JobAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-			Op:      db.BackupOperation,
-			Owner:   params.Owner,
-			JobUUID: id,
+			Op:           db.BackupOperation,
+			Owner:        params.Owner,
+			JobUUID:      id,
+			TaskUUIDChan: taskChan,
+		}
+
+		var taskUUID string = <-taskChan
```
jhunt: `:=` again

supervisor/jobs_api.go
```diff
@@ -142,13 +142,23 @@ func (self JobAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		if taskUUID == "" {
+			w.WriteHeader(500)
+			JSONLiteral(w, fmt.Sprintf(`{"ok":"error"}`))
+		} else {
+			w.WriteHeader(200)
+			JSONLiteral(w, fmt.Sprintf(`{"ok":"scheduled","task_uuid":"%s"}`, taskUUID))
```
jhunt: as above, check on usage of `JSONLiteral` elsewhere and either fix or ticket.  I want to avoid use of JSONLiteral that could run afoul of quoting antics (this one doesn't but sets precedence for fmt.Sprintf-driven "literals")

jhunt:
Structure of the proposed changes looks great.  Minor issues with style, and a question on JSONLiteral usage remain.

jhunt:
Also, for some reason Travis is borked?  Can you restart the builds to see if there are any real problems?  Definitely looks like a problem on Travis' end:

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

thomasmmitchell:
Not a travis problem. A test suite problem. The tests are failing because the JSON returned from the API call changed, and I'm fixing that... but theres an issue where a failure causes it to hang indefinitely. I can replicate it locally.

supervisor/archives_api.go
thomasmmitchell: JSONLiteral is used like this in all the supervisor libraries. If you really feel there needs to be a change, its probably outside of the scope of this ticket, yeah.

cmd/shield/main.go
geofffranks: I think these comments are all out of date, but i'm not sure. OK isn't called when options.Raw is specified here.. so the issue would be on the RawJSON call passing in task_uuid of an empty string variable.. which should be fine, since it's just `""`.

supervisor/jobs_api.go
```diff
@@ -142,13 +142,24 @@ func (self JobAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			Owner:        params.Owner,
+			JobUUID:      id,
+			TaskUUIDChan: taskChan,
+		}
+
+		taskUUID := <-taskChan
+
+		if taskUUID == "" {
+			w.WriteHeader(500)
+			JSONLiteral(w, fmt.Sprintf(`{"ok":"error"}`))
```
geofffranks: Why is this `{"ok":"error"}`? shouldn't we return something more like `{"error":"here is the reason for the error"}`?

supervisor/archives_api.go
```diff
@@ -92,17 +92,25 @@ func (self ArchiveAPI) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			RestoreKey:   archive.StoreKey,
+			TaskUUIDChan: taskChan,
 		}
 
-		w.WriteHeader(200)
-		JSONLiteral(w, fmt.Sprintf(`{"ok":"scheduled"}`))
+		taskUUID := <-taskChan
+		if taskUUID == "" {
+			w.WriteHeader(500)
+			JSONLiteral(w, fmt.Sprintf(`{"ok":"error"}`))
```
geofffranks: Why is this `{"ok":"error"}`? shouldn't we return something more like `{"error":"here is the reason for the error"}`?

supervisor/supervisor.go
```diff
@@ -125,15 +125,23 @@ func (s *Supervisor) ScheduleAdhoc(a *db.Task) {
 		// expect a JobUUID to move to the schedq Immediately
 		for _, job := range s.jobq {
 			if !uuid.Equal(job.UUID, a.JobUUID) {
+
```
geofffranks: Why the extra newline? formatting library woes?

supervisor/supervisor.go
```diff
@@ -125,15 +125,23 @@ func (s *Supervisor) ScheduleAdhoc(a *db.Task) {
...
 				continue
 			}
 
 			log.Infof("scheduling immediate (ad hoc) execution of job %s [%s]", job.Name, job.UUID)
 			task, err := s.Database.CreateBackupTask(a.Owner, job)
 			if err != nil {
 				log.Errorf("job -> task conversion / database update failed: %s", err)
+				if a.TaskUUIDChan != nil {
+					a.TaskUUIDChan <- ""
```
geofffranks: should we pass the error here back through the channel, so the API can return more meaningful error messages on failure? is that more work than it's worth?

supervisor/supervisor.go
```diff
@@ -151,8 +159,15 @@ func (s *Supervisor) ScheduleAdhoc(a *db.Task) {
 		task, err := s.Database.CreateRestoreTask(a.Owner, archive, target)
 		if err != nil {
 			log.Errorf("restore task database creation failed: %s", err)
+			if a.TaskUUIDChan != nil {
+				a.TaskUUIDChan <- ""
+			}
```
geofffranks: Same question here about the error message passing

supervisor/jobs_api.go
thomasmmitchell: I can change it to that. I just opted to keep the "ok" key in the JSON intact for ease of parsing. If somebody actually has an opinion, I'll go with that. I'll make that change.

supervisor/archives_api.go
thomasmmitchell: Answered above.

supervisor/supervisor.go
thomasmmitchell: Either that or a side effect of me moving code around. Maybe just to bother a well-intending code reviewer. The world may never know.

supervisor/supervisor.go
thomasmmitchell: It would take a little rearchitecting, but its not work I'm afraid of doing. The question, though, is what the end-user should see vs what the person trying to find the bug should see - i.e. what goes in an http response vs an error log.

cmd/shield/main.go
jhunt: The original issue was the compound conditional checking for "do we have a task UUID _and_ do they want raw output?"

If so, we were providing raw output; otherwise we were printing human-friendly output.

The complaint was that if the task UUID is "", the consequent of the conditional is skipped, and it reads like this: "well, task UUID was empty, so I guess print human-readable output.  Granted, `Ok` will make that JSON, but it would be JSON without the `task_uuid` attribute, which would be confusing.  Also, if someone opted to print some additional messaging (like `fmt.Printf("but we found no task UUID!\n")` it would break JSON parsers.

IIRC, @thomasmmitchell already fixed the code.

karampok:
nice, the changes in the API is exactly we wanted.
thx for your work!

PR #188, opened 2016-06-23
correct a small bug during a post restore call

karampok:
I am getting 
```
--$ curl -XPOST  -H "X-Shield-Token: LETMEIN"  http://localhost:8181/v1/archive/58831f9a-1b02-4231-ab92-13e6adfacb47/restore
{"error":"a631f230-e68e-4766-82b3-9915a8b2b449"}
```
but the task id is valid and the job is finished.


jhunt:
Thanks!

PR #189, opened 2016-06-23
added tests around api responses when queuing restores

thomasmmitchell:
Tests that nobody asked for


thomasmmitchell:
ugh, noticed some paste residue

PR #193, opened 2016-06-24
implemented exact name matching in API

### LINKED ISSUE #112: Allow cli not to use regex when trying to list object
Currently if we do list store its a regex for name and cli returns multiple results, can we have `shield --raw list store s3` only match exact name vs regex matching with %

Log example for command above.
```
2016-02-22 15:29:16.647966763 +0000 UTC shieldd: DEBUG: Parameters: [%s3%]
2016-02-22 15:29:23.463956501 +0000 UTC shieldd: DEBUG: Executing SQL:
            SELECT s.uuid, s.name, s.summary, s.plugin, s.endpoint, -1 AS n
                FROM stores s
                WHERE s.uuid = s.uuid AND s.name LIKE $1
                ORDER BY s.name, s.uuid ASC
```

thomasmmitchell:
also cli support and tests for api

Fixes #112

### DIFF
--- a/supervisor/jobs_api_test.go
+++ b/supervisor/jobs_api_test.go
@@ -39,109 +39,115 @@ var _ = Describe("/v1/jobs API", func() {
 	PG_S3_WEEKLY := `ea5eec7a-20b8-40ef-9868-cb4c54133018`
 	REDIS_S3_DAILY := `9f42b026-8698-4150-b508-561a73031a47`
 	REDIS_S3_WEEKLY := `4259dd16-0fad-478b-815c-9a9a7d209f56`
+	REDIS_S3_WEEKLY_PLUS := `56cb2511-3286-449b-970e-cb7f21db145c`
 
-	BeforeEach(func() {
-		data, err := Database(
-
-			// TARGETS
-			`INSERT INTO targets (uuid, name, summary, agent, plugin, endpoint) VALUES
-				("`+TARGET_REDIS+`",
+	databaseEntries := []string{
+		// TARGETS
+		`INSERT INTO targets (uuid, name, summary, agent, plugin, endpoint) VALUES
+				("` + TARGET_REDIS + `",
 				 "redis-shared",
 				 "Shared Redis services for CF",
 				 "10.11.22.33:4455",
 				 "redis",
 				 "<<redis-configuration>>")`,
-			`INSERT INTO targets (uuid, name, summary, agent, plugin, endpoint) VALUES
-				("`+TARGET_PG+`",
+		`INSERT INTO targets (uuid, name, summary, agent, plugin, endpoint) VALUES
+				("` + TARGET_PG + `",
 				 "pg1",
 				 "Test Postgres Service",
 				 "10.11.22.33:4455",
 				 "pg",
 				 "<<pg-configuration>>")`,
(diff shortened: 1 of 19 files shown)


supervisor/jobs_api_test.go
```diff
@@ -524,7 +525,38 @@ var _ = Describe("/v1/jobs API", func() {
...
 				}
 			]`))
-		Ω(res.Code).Should(Equal(200))
+			Ω(res.Code).Should(Equal(200))
+		})
+	})
+
+	Context("With exact matching", func() {
+		It("should not retrieve jobs with only partially matching names", func() {
+			res := GET(API, "/v1/jobs?name=Redis+Daily+Backup&exact=t")
```
jhunt: Does this really test the "exact" nature of the query?  You almost need another fixture named `Redis Daily Backup - Extra Stuff`...

supervisor/retention_api_test.go
```diff
@@ -85,6 +85,29 @@ var _ = Describe("HTTP API /v1/retention", func() {
...
+					"expires" : 1209600
+				}
+			]`))
+			Ω(res.Code).Should(Equal(200))
+		})
+	})
+
+	Context("With exact matching", func() {
+		It("should not retrieve policies with only partially matching names", func() {
+			res := GET(API, "/v1/retention?name=sho&exact=t")
```
jhunt: Can you also include a positive single match (prefix, like the Redis job above?), possibly with some new fixtures?

supervisor/schedules_api_test.go
```diff
@@ -87,6 +87,29 @@ var _ = Describe("HTTP API /v1/schedule", func() {
...
+					"when"    : "daily at 11:24pm"
+				}
+			]`))
+			Ω(res.Code).Should(Equal(200))
+		})
+	})
+
+	Context("With exact matching", func() {
+		It("should not retrieve schedules with only partially matching names", func() {
+			res := GET(API, "/v1/schedules?name=dai&exact=t")
```
jhunt: Again, a positive single (prefix?) match test would be nice.

supervisor/stores_api_test.go
```diff
@@ -90,6 +90,30 @@ var _ = Describe("/v1/stores API", func() {
...
+					"endpoint" : "<<redis-configuration>>"
+				}
+			]`))
+			Ω(res.Code).Should(Equal(200))
+		})
+	})
+
+	Context("With exact matching", func() {
+		It("should not retrieve stores with only partially matching names", func() {
+			res := GET(API, "/v1/stores?name=red&exact=t")
```
jhunt: Once more, with prefix single match!

supervisor/targets_api_test.go
```diff
@@ -94,6 +94,31 @@ var _ = Describe("/v1/targets API", func() {
...
+					"endpoint" : "<<redis-configuration>>"
+				}
+			]`))
+			Ω(res.Code).Should(Equal(200))
+		})
+	})
+
+	Context("With exact matching", func() {
+		It("should not retrieve targets with only partially matching names", func() {
+			res := GET(API, "/v1/targets?name=red&exact=t")
```
jhunt: Here too.

Issue #184, opened 2016-06-17
Mysql as a back database for SHIELD


karampok:
Greeting, 
Longterm we would like to have the alternative to use mysql as database. It will simplify the setup/deployments given most of our services are running on mysql (nevertheless with bosh this is not a problem). Anyway, long story sort,  this can be an external requirement we will need to handle.

Currently SHIELD support PostgreSQL, and SQLite. In the code, I saw that you are using directly SQL queries, therefore due to semantics (uuid, $1 bindvars) mysql cannot be supported out of the box.
A future work which will refactor the database interaction part using a framework like
https://github.com/go-gorp/gorp
https://github.com/jmoiron/sqlx
https://github.com/jinzhu/gorm
in order to make SHIELD  database agnostic would be something that you would like to see/work on?

Thx

jhunt:
SHIELD only supports PostgreSQL as a database backend.   We utilize SQLite to avoid requiring a spinning PostgreSQL process while doing tests, but do not support (nor recommend) its use in real-world deployments.

Refactoring the (currently functional) database layer to support arbitrary database backends like MySQL, Oracle, or even MS SQL server brings additional complexity to the codebase for no real gain to the functionality of SHIELD itself.  To do this properly, we would then also need to modify the SHIELD BOSH release to allow selection of a database backend, as well as jobs for alternate backends (at least MySQL).

All of this added complexity brings with it additional support and testing overhead.  If we say we support MySQL as a backend (via an ORM or otherwise) we will need to vet the releases against both PostgreSQL and MySQL, doubling the work involved in a release.

Given that the current state (SHIELD using explicitly PostgreSQL) has no bearing on its ability to _back up_ MySQL target systems, I see little gain in reworking a perfectly functional (and well-tested) layer of the code.

Are you unable to use the `postgres` job that stands up a database inside the BOSH deployment?

karampok:
Yes, i can understand that refactoring the database layer is a big task without obvious benefit to SHIELD itself. If we introduce changes, this should be only by picking up a good  ORM which will provide the abstraction in such extent that there will not be a need to validate SHIELD everytime against  different store backend (if possible of course). 

To your question, yes we will be unable to use postgres as defined in BOSH deployment. We will be required eventually  to use mysql db as back store, because all other  applications in our are setup are doing so.

Saying that, will you be willing to accept a pull request from our side  towards that direction?
Thanks!

jhunt:
Understand completely.

[...]

PR #198, opened 2016-07-05
support MySQL (#184)

gogolok:
Based on the recent sqlx addition, we can now support MySQL/MariaDB as a backend.

- introduce MySQL driver github.com/go-sql-driver/mysql
- MySQL/MariaDB uses another schema than PostgreSQL. We moved the schema to different packages with same interfaces. #184 suggests we put it into the same functions. Can someone elaborate on that idea?
- the db.Connect method  appends sql_mode POSTGRESQL when not set to ensure that the operator "||" can be used to concatenate strings
- Can you clarify why the type 'text' gets used for certain status columns insteaf of varchar(X) for example? MySQL has trouble supporting `text DEFAULT 'default-value'`. This would simplify the differences in the schema.


jhunt:
My concern with having the same logical schema in two different packages is that they will not be updated in parallel.  Perhaps this is just a pothole we have to deal with based on how Go works, but my initial thought was to switch inside the migration function, and run different queries based on targeted database engine.  i.e. for Postgres, do X, for MySQL, do Y.  This also means that we don't run the risk of having different schema versions for different backends when (for example) we have to tweak the MySQL schema only, and we bump the schema version -- does the PostgreSQL schema contain a weird, empty schema function for that version (since it was unaffected?)

TEXT was used to avoid placing arbitrary limits on column fields.  Switching to VARCHAR should be acceptable.

karampok:
I agree that having two different schemas to be maintained in parallel,  is not ideal. 

Two different schemas came from the fact that we have
1. UUID types that exists in PostgreSQL, and not in MySQL (+ PRIMARY KEY)
2. the `text DEFAULT 'default-value'`, which is also not supported in MySQL
3. other variables (e.g. the InnoDB engine) that are only MySQL related

We thought it would be better option to keep the PostgreSQL schema/current workflow intact, and develop in parallel and independently introducing minimal changes. 

Do you think there will be cases when we will have to tweak only one of database schemas?
Different schema version per database type could also be an option.

On the other side, If we would like to merge them:
for the number 2 we could change it in general,  if that is ok with you,
for the number 3 we could enforce it in the DSN
for the UUID type is the question, we could do a custom function that translates the UUID to varchar(36) before the sql query, not perfect though. What do you think?

jhunt:
I understand the rationale for two different schemas - different database engines need different things.  (This is one of the reasons we standardized on PostgreSQL to begin with; ORMs only net you so much).

My concern is that by maintaining the SQL in two different files, it becomes difficult to compare them for equivalence.  Can we do something like this instead?

```
package schema

// ...

func (s v1Schema) Deploy(db *DB) error {
  var err error
  switch db.Type {
  case "mysql":
    err = db.Exec(`CREATE TABLE schema_info (version INTEGER) ENGINE = InnoDB`)
  case "postgres":
    err = db.Exec(`CREATE TABLE schema_info (version INTEGER)`)
  default:
    err = fmt.Errorf("unsupported database type '%s'", db.Type)
  }
  if err != nil {
    return err
  } 
  // ...
}
```

karampok:
okay, so basically you suggest that we have indeed two set of SQL statements if  needed (schema/migrations per database type), but located always in the same go file so we can easily compare and maintain them.  Yes this sound better for sure, we can also have single case statement if SQL statement applies to both types. @gogolok do you agree?

gogolok:
Okay, will try that approach

gogolok:
We've updated the sources:

- The MySQL schemas are within the schema version files.
- The special requirements on the DSN (sql_mode POSTRGESQL && default_storage_engine=INNODB) have been removed from the db.Connect method to minimize the code changes. The shield-boshrelease will have a working DSN example.

Do the changes look better now? Anything else missing?

jhunt:
Looks pretty good. I will run some more involved tests later to verify that Postgres stuff still works (nothing leads me to believe it _won't_, just want to verify) and then get this merged.

karampok:
If it looks ok, could we merge?
Thx!

jhunt:
Still evaluating.

geofffranks:
I have some concerns with the length of some of your VARCHARs. Limiting them at 255, when they're going to contain the same data as fields from other tables that are TEXTs is going to cause strange issues down the road. 

For example, the `target.endpoint` column is TEXT, and gets copied into `tasks.target_endpoint`, which is a max of 255 characters. Same for the other endpoint, plugin agent, and restore_key related columns. How would you feel about VARCHAR(65535) for those?

gogolok:
Good point. Will change it to VARCHAR(65535)

gogolok:
Ok, MySQL/MariaDB has really some strange limitations. 👎

Although VARCHAR supports the maximum size at 65535 characters, the actual maximum value depends on other columns in the table and character set. Maximum row size is 65535 bytes in MySQL that shared among all columns in the table (except for BLOB and TEXT). Encoding further limits the maximum length of VARCHAR.

jhunt:
Why not use TEXT?

geofffranks:
IIRC mysql didn't allow for TEXT + defaults. Though perhaps no default/ null would get translated properly to an empty string when pulled out via the go library?

gogolok:
MySQL TEXT can't have a default value.

If you don't have enabled strict mode, you can use TEXT DEFAULT ''. Then we would only have an issue with `ALTER TABLE archives ADD COLUMN status TEXT DEFAULT 'valid'`

jhunt:
Columns like `status` don't need to be TEXT; they are more enums than anything else.

Things like `endpoint` on the other hand *do* need to be TEXT, because they are open-ended JSON configuration provided by end users.

PR #202, opened 2016-07-15
allow only a single task per target

### LINKED ISSUE #200: Allow only a single task per target
Currently it seems that we can trigger a restore job and a backup job for the same target.  Is that correct?
It would probably make sense to enforce that only a unique task is running per target.
Something like   “if this target has a task in the task database, hold off”.

gogolok:
This is an attempt to fix #200 

Since task scheduling is not done in parallel, we can check prior to
starting a task whether it is allowed to run that task.

A task may not run if a task with same target uuid is running. This is
useful if a restore operation is currently in progress and another
backup should be initiated.

### DIFF
--- a/db/tasks.go
+++ b/db/tasks.go
@@ -247,6 +247,21 @@ func (db *DB) CreatePurgeTask(owner string, archive *Archive, agent string) (*Ta
 	return db.GetTask(id)
 }
 
+func (db *DB) IsTaskRunnable(task *Task) (bool, error) {
+	r, err := db.Query(`
+		SELECT uuid FROM tasks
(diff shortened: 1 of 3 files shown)


jhunt:
Looks a lot better to me.  I'd like @thomasmmitchell or @geofffranks to weigh in as well before merging.

thomasmmitchell:
lgtm

gogolok:
Found an issue with the code. Will check and attampt to fix tomorrow.

gogolok:
I've fixed my mistake. We have to adjust the index after deleting a task from scheqd.

PR #207, opened 2016-09-09
pg_bindir

wayneeseguin:
Enabling specification of pg_bindir similar to already configurable items in the plugin configuration.
This gets rid of the hard-coded path assumption while leaving a default that isn't pointing to a specific version of postgres but simply assumes a 'postgres' package as default. Most important point is to be able to configure where the pg binaries bin dir is.

### DIFF
--- a/plugin/postgres/plugin.go
+++ b/plugin/postgres/plugin.go
@@ -21,9 +21,9 @@
 //        "pg_user":"username-for-postgres",
 //        "pg_password":"password-for-above-user",
 //        "pg_host":"hostname-or-ip-of-pg-server",
-//        "pg_port":"port-above-pg-server-listens-on",   # optional, defaults to '5432'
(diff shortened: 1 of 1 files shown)


gogolok:
General question: Why do we need to specify the path to the psql/... binary? Shouldn't the bosh-release setup the env accordingly e.g.?

jhunt:
@gogolok not everyone names their postgres package `postgres`.

geofffranks:
This brings back into play the `pg_bindir` option, which as far as i can tell was removed to simplify the need for specifying all the time (pre-defaults framework). My only concern would be that existing installations have `pg_bindir` specified still, but pointing to a bad path. Once they upgrade to this code, their jobs may start failing. I'll put a note in the release notes about this, but otherwise, 🚢 it.

PR #212, opened 2016-10-20
fix db sql statement to update purge reason in archives

ampersand8:
This will fix the problem, that the purge_reason never gets set when purging,
because of an invalid sql statement (subselect on same table as updating).

### DIFF
--- a/db/archives.go
+++ b/db/archives.go
@@ -228,10 +228,9 @@ func (db *DB) PurgeArchive(id uuid.UUID) error {
 		return fmt.Errorf("Invalid attempt to purge a 'valid' archive detected")
 	}
 
-	err = db.Exec(`UPDATE archives SET purge_reason =
-                       (SELECT status FROM archives WHERE uuid = ?)
-                       WHERE uuid = ?
-                  `, id.String(), id.String())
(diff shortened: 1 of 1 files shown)


jhunt:
Has this been tested against the other support database platforms (postgres / sqlite)?

gogolok:
I will hopefully test it this or next week against all supported database platforms.

jhunt:
I'm going to hold off on merging then, until that testing is completed and all platforms are happy.

ampersand8:
Indeed I previously only checked on our environment (mysql). I now updated the query, which hopefully works on all supported database platforms.

gogolok:
I've tested it on the different platforms. Works for me.

I'm wondering whether the update statement is just:
```sql
UPDATE archives SET purge_reason = status WHERE uuid = ?
```

ampersand8:
Yeah, I completely agree. Your suggested solution is much simpler and seems to work on mysql, postgres and sqlite.

jhunt:
Can we get an updated commit with the simplified SQL query?

jhunt:
Fixed in 5746cc6, independently.

PR #214, opened 2016-10-27
Fixed `fs` target plugin output to `/dev/stdout` as expected

bgandon:
The `fs` target plugin stores its backup into the `/dev/st0` file (the bsdtar default) instead of `/dev/stdout`, as expected by SHIELDS.

As a result, the files in `{base_dir}/yyyy/mm/dd/HH/MM/SS/<uuid>` are always 14 bytes long bzip2 files (which actually encode an empty content).

There's a workaround for this issue, which is to set a `bsdtar: /var/vcap/packages/bsdtar/bin/bsdtar -f /dev/stdout` config for the `fs` target plugin.

In this PR, I submit a permanent fix.


jhunt:
Thanks!

PR #216, opened 2016-10-30
Bail out if no data is received by shield-pipe

### LINKED ISSUE #215: Trigger failures in store plugins when no data is received
Add catch-all functionality to cause store plugins to error out if they detect no data inbound from stdin to actually store on the filesystem.

This will help counteract situations where target plugins exit non-zero but do not provide any data to be backed up to stdout

jhunt:
Fixes #215

These modifications to shield-pipe correctly detect the case where a
plugin is not sending data to the other side.  For backup operations,
this catches bad backups.  The restore-side is dubious, but applying the
patch there as well makes for some nice symmetry, and shouldn't hurt
anything.

The `head -c1` command grabs up to the first byte out of the stream and
writes it to a temporary file.  This should help us to keep the disk
impact as alow as possible, but it does mean that we know rely on disk
to be writable (at least `/tmp`)

cc @geofffranks 

Please don't merge this until Geoff has had a chance to review in light of #215

### DIFF
--- a/bin/shield-pipe
+++ b/bin/shield-pipe
@@ -98,9 +98,21 @@ case ${SHIELD_OP} in
 	validate STORE  ${SHIELD_STORE_PLUGIN}  "${SHIELD_STORE_ENDPOINT}"
 
 	header "Running backup task (using bzip2 compression)"
+
+	PULSE=$(mktemp --tmpdir shield-pipe.XXXXX)
+	trap "rm -f ${PULSE}" QUIT TERM INT
+
 	set -o pipefail
-	${SHIELD_TARGET_PLUGIN} backup -e "${SHIELD_TARGET_ENDPOINT}" | bzip2 | \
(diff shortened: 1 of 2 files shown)


thomasmmitchell:
lgtm. only thing is that could you possibly store that character into a bash string with `>>>` rather than using a tmpfile, or is this just something for shell compatibility?

jhunt:
If it works, I like that idea even better.  I will give it a shot

jhunt:
Sadly, the only way to make this work without fs access is even _more_ complicated than tee-ing in the `head -c1` call, so I'm inclined to leave it as-is.  Any site that is strictly concerned about use of the filesystem (in a readonly scenario) can set TMPDIR to a tmpfs backed by RAM, not disk.

jhunt:
merged/rebased/pushed.  Now, _that's_ conflict-resolution.

Issue #209, opened 2016-10-10
Issues when restoring non 'postgres' pg database


bodymindarts:
We had some issues while testing the restoration of a BOSH.
Goal was to recreate a previously backed up BOSH on a cleanly deployed BOSH.

The issues we faced were:

The configuration "pg_database": "bosh" was not respected when performing a restoration. Instead of restoring the bosh database all tables were restored into the postgres database.
work around:
Don't configure "pg_database" so that all databases are backed up.

With a backup of all databases it was still not possible to restore database bosh while connections were present.
work around:
sudo monit stop all to drop the connections prior to restoration.

Once restoration was completed the ROLE boshdb had the attribute NOLOGIN.
work around:
ALTER ROLE boshdb LOGIN CREATEDB; to give the role the same permissions it had previously.

Though it is possible to perform this recreation there are some manual steps involved (dropping connections and altering the role). Is there a way we could streamline this process and have the plugin handle it better?

jhunt:
I was under the impression that all of these manual workarounds were already supported in the plugin. Let me try to set up a lab locally to reproduce this and we'll see what's up

jhunt:
Question: did you backup *all* of the databases but hope to only _restore_ a subset?  That use case is not supported; essentially, the `backup` task honors the `pg_database` when it picks what to backup, but the `restore` task just restores what's in the SQL file made by the `backup` task.

bodymindarts:
We tried both backup and restore of _just one_ database. That didn't work so we moved to backing up and restoring _all_ databases which did work except for the issue with the connections and user roles.

johnlonganecker:
The `psql -d postgres` is hard coded:

https://github.com/starkandwayne/shield/blob/master/plugin/postgres/plugin.go#L204

If you are only backing up one specific database I believe the backup file doesn't contain the name of that database. 

When you are backing up/restoring all the databases the backup file explicitly declares the database name and overrides the `-d postgres`

I am 99% sure this is the problem and will double check and submit a PR.

PR #245, opened 2017-01-28
fix postgres single database backup bug

johnlonganecker:
When backing up one database the postgres plugin will restore that to the database with the name `postgres` even if the database name was something different.

When performing a backup of all databases this bug does not show up.

This fixes issue #209


cweibel:
The current behavior is attempting to overwrite the `postgres` database during restore because that is the database that is initially connected to?

PR #246, opened 2017-01-30
remove usage of deprecated commands after creating new things

johnlonganecker:
After running `create-*` commands you would receive a message:

`The alias show target is deprecated in favor of target`

I made sure we were using the latest `show/list` commands after we created a new policy, schedule, target, store, etc.


jhunt:
Thanks!

PR #250, opened 2017-02-02
Make mysql binary directory configurable

johnlonganecker:
If you are running the mysql plugin right now you are forced to put the plugin in `/var/vcap/packages/shield-mysql/bin/` 

The postgres plugin allows you to specify a binary directory, so I added that feature to the mysql plugin.

If left undefined it defaults to `/var/vcap/packages/shield-mysql/bin/` this change will not break existing implementations.

### DIFF
--- a/plugin/mysql/plugin.go
+++ b/plugin/mysql/plugin.go
@@ -24,7 +24,8 @@
 //        "mysql_port":"port-above-mysql-server-listens-on",
 //        "mysql_read_replica":"hostname-or-ip-of-mysql-replica-server",  #OPTIONAL
 //        "mysql_database": "your-database-name",  #OPTIONAL
-//        "mysql_options": "mysqldump-specific-options" #OPTIONAL
+//       	"mysql_options": "mysqldump-specific-options", #OPTIONAL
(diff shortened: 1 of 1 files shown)


plugin/mysql/plugin.go
```diff
@@ -24,7 +24,8 @@
 //        "mysql_port":"port-above-mysql-server-listens-on",
 //        "mysql_read_replica":"hostname-or-ip-of-mysql-replica-server",  #OPTIONAL
 //        "mysql_database": "your-database-name",  #OPTIONAL
-//        "mysql_options": "mysqldump-specific-options" #OPTIONAL
+//       	"mysql_options": "mysqldump-specific-options", #OPTIONAL
+//        "mysql_bindir": "/mysql/binary/directory" #OPTIONAL
```
lnguyen: 👎

norman-abramovitz:
The code looks good, but I would like to suggest that you make a variable for these hard coded values.

johnlonganecker:
I think that is a good idea norm, maybe we can create another issue for that? the postgres and mysql plugins both do it this way.

I think this PR is good to merge?

norman-abramovitz:
Agree

Issue #240, opened 2017-01-06
Support alternative exit codes for mysql plugin


janaurka:
Hey

This is a feature request.

It would be great if the `mysql` plugin would support variable `exit codes`. Currently it will fail on an `exit code != 0`, which should be fine most of the time.

The problem is that you (as DB operator and in charge of backups) are not always capable of having a proper state in each `dbs`.  For example: We have a case where we have some `VIEWS` which belong to a not anymore existing user in a `dev` db. The Devs don’t really care, but `mysqldump` fails to backup the db.

That’s why I’ve created https://github.com/starkandwayne/shield/pull/239 . Unfortunately this solves only one part of the puzzle. With https://github.com/starkandwayne/shield/pull/239 we’d be able to tell `mysqldump` to `--force` and ignore the erros, _but_ `mysqldump` then will escape with a `non-zero` exit code (in our case it’s `2`) which will be evaluated as a `failed backup` by `shield`.

`shield` comes prepared to respect other `exit codes` but the `mysql` plugin needs some changes to make this also possible there. I’ve started working on the implementation here: https://github.com/janaurka/shield/tree/exit_codes .

_Note:_ Next to the `exit codes` the `mysql` plugin needs to be refactored to use `ExecWithOptions` instead of `Exec` (which is already done in https://github.com/janaurka/shield/tree/exit_codes).

I experienced a problem with validating the `exit codes` which are expected as `[]int` like `[]int{0,2}`, which would allow `exit codes` `0` and `2`.

There is https://github.com/janaurka/shield/blob/exit_codes/plugin/endpoint.go#L99 which allows `[]interface` and i’ve added https://github.com/janaurka/shield/blob/exit_codes/plugin/endpoint.go#L112 which would allow to set a `default` (here it would be `[]int{0}`), but the `go compiler` complains about:

```
(type []int) as type []interface {} in argument to endpoint.ArrayValueDefault
```

Which is probably because of https://github.com/golang/go/wiki/InterfaceSlice. (But my `go` skills are to immature to solve the problem atm :( .

Maybe if you point me towards the right direction, I’d be able to fully implement this on my own. I’ve also tried to ask on the [Slack channel](https://cloudfoundry.slack.com/archives/shield/p1483622488000008).

cheers
j

jhunt:
I'm not really sure what the real nature of the feature request is.  Does it make more sense to just not enforce exit code checking if the operator so wishes?  I.e., something like:

```
{
  "mysql_user":     "username",
[...]

PR #253, opened 2017-02-08
Add functionality to ignore the ExitStatus

janaurka:
This is an implementation for #240 

This commit introduces CheckExitStatus (default: true) for plugin
execs. If CheckExitStatus is set to false, the process will ignore
the exit status and will set the action to sucess even when the
invoked backup mechanism exited with an exit code != 0.

The support is currently only built into the mysql plugin.

Unfortunately this is implemented using the strings "true" or
"false" because of my lacking go skills.

But it works. I have tested it using a DEV shield-boshdeployment
release.

Note that I have tried to use `boolean` values for `CheckExitCode`. The problem I did run in was, that if the `struct` consist of `bool`, then `go` will evaluate it to `false` (go zero values) and there is no way to figure out if the code was set and we need to know since it’s optional.

A workaround would be to work with pointers `CheckExitCode *bool`, since an empty pointer always is `nil` and would allow something like:

```
if opts.CheckExitCode != nil {
  CheckExitCode = opts.CheckExitCode
}
```
But this somehow did not work (probably because I am too stupid for it :) ). So consider it as a solutions, but not as a really nice one...

### DIFF
--- a/plugin/exec.go
+++ b/plugin/exec.go
@@ -12,12 +12,17 @@ const NOPIPE = 0
 const STDIN = 1
 const STDOUT = 2
 
+var (
+	DefaultCheckExitStatus = "true"
+)
+
 type ExecOptions struct {
-	Stdout   *os.File
-	Stdin    *os.File
-	Stderr   *os.File
-	Cmd      string
-	ExpectRC []int
+	Stdout          *os.File
+	Stdin           *os.File
+	Stderr          *os.File
+	Cmd             string
+	ExpectRC        []int
+	CheckExitStatus string
 }
 
 func ExecWithOptions(opts ExecOptions) error {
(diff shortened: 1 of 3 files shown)


janaurka:
@jhunt I’ve thought about this and the only viable way I see right now would be to have something like a `general` option which can be used for the plugins. Then this could be patched into `Exec` (always check) and `ExecWithOptions` (make it configurable).

For an example, the mysql plugin could get configured as:

```
{
  "general_check_exit_code": "true",
  "mysql_user": "root",
  "mysql_password": "not-the-password",
  "mysql_host": "127.0.0.1",
  "mysql_port": "3306"
}
```

Then the `Run(p Plugin)` in `plugin/plugin.go` would neet to be enhanced to parse additional info into `opts` which then should make this generic and usable for all plugins. What do you thin!

jhunt:
Another option is to lift the functionality out of the plugins altogether, and have a flag on the target / store to "ignore non-zero plugin exit codes".  Mechanism-wise, we should be able to just let `shield-pipe` know about the new flag, communicate it through the agent/worker channel, and it should just work.

Unfortunately, these types of changes require uniform upgrades, and we have lots of users with distributed setups who may not be able to upgrade plugins right away.

We also have changes in the works that would more tightly couple plugin-agent-core interaction, to allow things like front-end enumeration of available plugins and available options, so I'd really like to get those done before we make other changes to the remote side of the architecture.

janaurka:
I‘ll close this for now since there are big changes coming and i currently do not have the time to invest here. :(

PR #254, opened 2017-02-13
moved shield docker compose from its own repo into the shield repo

johnlonganecker:
A fully running shield setup inside docker containers, with a tutorial on how to use shield.


jhunt:
Thanks!

PR #259, opened 2017-03-05
Document the normal (non raw) mode for 'shield restore' cli

bgandon:
Hi guys,

The `shield help restore` was incomplete and actually misleading about the way `shield restore` was supposed to be used.
What I did here in this PR is document the normal (non `--raw`) mode for the `shield restore` to be clearer.

Cheers


jhunt:
Thanks!

PR #280, opened 2017-05-07
Fix API library leaking connections

frodenas:
I've been observing that when using the API library, connections are leaked (see https://github.com/cloudfoundry-community/shield_exporter/issues/3). This PR fixes this by:

* SecureBackendURI function reads (and discards) the ping requests to avoid leaking (`ESTABLISHED`) connections
* Disable Keep Alives, as connections are not reused
* Set a connection timeout
* If environment has Proxy settings, use them when setting up the connection

### DIFF
--- a/api/config.go
+++ b/api/config.go
@@ -5,12 +5,14 @@ import (
 	"encoding/base64"
 	"fmt"
 	"gopkg.in/yaml.v2"
+	"io"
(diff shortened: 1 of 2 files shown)


jhunt:
In the future, I'd prefer to see the timeout / proxy settings as a separate PR from the bug fix PR.

PR #292, opened 2017-06-16
create-* actions have --update-if-exists flag

drnic:
`create-store`/`create-target`/`create-schedule`/`create-policy` commands now support an `--update-if-exists` flag which will find the first item by name and updated it; rather than create a duplicate.

Example:

    cat <<EOF | shield --raw --update-if-exists create-schedule
    {
      "name":"daily",
      "when":"daily 3am"
    }
    EOF


jhunt:
Thanks!

PR #296, opened 2017-07-04
v6.x - S3 plugins : fix Radosgw bug + default scheme S3 client

nabbar:
Hi,

I have made this update of branch v6.x to allow specify the default scheme for S3 plugins and update the minio-go lib to fix bug with radosGW S3.

Details of updates : 
* Transform the hard-coded secure mode false in S3client as a config options to specify a secure S3 connection (default value is set to false for compat.)
* Fix S3 RadosGW bug ([#629](https://github.com/minio/minio-go/issues/629) & [#630](https://github.com/minio/minio-go/pull/630)) : 
  * Bump library github.com/minio/minio-go to release 2.1.0 in Godeps/Godeps.json
  * Update folder vendor/github.com/minio/minio-go (using godep tools)

(Same as PR #297 but for v6.x branch)


jhunt:
Before this gets merged, @geofffranks and @thomasmmitchell both need to weigh in.  I'm wary of merging it because IIRC, we use a forked copy of the minio library from before they switched the disposition of the `ssl bool` param to NewV2 from one meaning to another.  I.e. it used to mean "insecure", and then they changed it to mean "secure", so the `false` in the code was to force SSL at all times.  Changing this may have serious ramifications for users of the `s3` plugin.

thomasmmitchell:
My main thing would've been that we need to keep the S&W fork of the minio library for the scality plugin (because of something off-spec with packet sizes in multipart uploads). Which all checks out in this PR, so I'm happy.

I know there's something about the ssl bool being flipped that Geoff has told me about... but I don't think I've ever verified it, and should probably wait until he's around tomorrow to look into that aspect of it...

PR #298, opened 2017-07-07
Update consul plugin to use consul-snapshot

bodymindarts:
Since Consul 0.7.1 the command `consul snapshot` has been available to backup and restore all of Consuls entities to a local file.
This PR updates the consul plugin to make use of the native consul snapshot command for backup and restore.

### DIFF
--- a/plugin/consul/plugin.go
+++ b/plugin/consul/plugin.go
@@ -16,16 +16,17 @@
 // endpoint JSON should look something like this:
 //
 //    {
-//        "host":"consul-endpoint",          # optional - can also be prefixed with http:// or https://
-//        "skip_ssl_validation":false        # optional
-//        "username":"basic-auth-username",  # optional
-//        "password":"basic-auth-password"   # optional
+//        "address":"consul.service.consul:8500",                    # optional - can also be prefixed with http:// or https://
+//        "ca-path":"/var/vcap/jobs/consul/consul/ca.cert",          # optional - required for connecting via https
+//        "client-cert":"/var/vcap/jobs/consul/consul/consul.cert",  # optional - required when verify_incoming is set to true
+//        "client-key":"/var/vcap/jobs/consul/consul/consul.key"     # optional - required when verify_incoming is set to true
 //    }
 //
 // Default Configuration
 //
 //    {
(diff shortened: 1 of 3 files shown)


plugin/consul/plugin.go
```diff
@@ -42,109 +43,153 @@
...
+func (p ConsulPlugin) Validate(endpoint plugin.ShieldEndpoint) error {
 	var (
 		s    string
-		b    bool
 		err  error
 		fail bool
 	)
 
-	s, err = endpoint.StringValueDefault("host", "")
+	s, err = endpoint.StringValueDefault("consul", "")
```
jhunt: Is this a backwards-breaking change?  Will existing configurations with a `host` key continue to work once this code is merged in?

plugin/consul/plugin.go
```diff
@@ -42,109 +43,153 @@
...
 	} else if s == "" {
-		ansi.Printf("@G{\u2717 host                  using default host @C{%s}}\n", DefaultHostPort)
+		ansi.Printf("@G{\u2713 consul}       using default consul @C{%s}\n", DefaultConsul)
 	} else {
-		ansi.Printf("@G{\u2713 host}                  @C{%s}\n", s)
+		ansi.Printf("@G{\u2713 consul}       @C{%s}\n", s)
 	}
 
-	b, err = endpoint.BooleanValueDefault("skip_ssl_validation", false)
+	s, err = endpoint.StringValueDefault("address", "")
```
jhunt: Is this also a backwards-breaking change?

bodymindarts:
You are right this is backwards incompatible. I'll add this code as a new plugin `consul-snapshot` and add a depreciation warning to the old one.

jhunt:
Why not just add a `use_snapshot` option to this plugin?

bodymindarts:
To keep the code cleaner.
The 2 plugins have different options and different mechanisms of execution and different outcomes.
Both from the code author and end-users perspective they are different.
I don't see a benefit of merging together these 2 plugins with a bunch of conditional branching code. Will make the code much less understandable.

jhunt:
A good justification.

I will go ahead and merge this PR.

PR #299, opened 2017-07-07
Add the ability to pass generic mongo options

skburgart:
Hey all,

We run Mongo with SSL enabled, but the mongo plugin does not have the ability to pass the `--ssl` flag to the mongo commands. Rather than add a flag specific to `ssl` we followed the example from the [mysql plugin](https://github.com/starkandwayne/shield/blob/master/plugin/mysql/plugin.go#L27) and added the `options` parameter that allows passing any flags.

#ultimatesoftware


jhunt:
Thanks!

PR #310, opened 2017-08-23
Cli refactor

thomasmmitchell:
The SHIELD CLI has grown considerably since its first iterations, and I've dealt with complaints from new people coming onto the codebase that it was difficult to navigate. While it worked great for us, who know where everything is in the project, it could be cumbersome if you hadn't built it up until this point and therefore didn't even know what to look for.

Larger changes were already going to need to be made to support 8.x SHIELD, so it seemed a fair time to do a major reorganization. Generally, the massive main.go file has been split into smaller files in different packages. For example, the commands `stores`, `store`, `create-store`, `edit-store`, and `delete-store` can be found in the `stores` package at `/cmd/shield/commands/stores`, and each function is bundled with its help information in its own respective file. All of the commands are imported and registered in the main package, which has been reduced to one file, `main.go`. Most of the helper functions that were used by the commands have been moved to the `cmd/shield/commands/internal` package, where all of the functions can use them.

Generally, there is no one true code style that is going to make everyone happy. There are tradeoffs to using a bunch of smaller files as opposed to a few really large ones, but I'm hoping that this is a design that we can agree to move forward with.

I generally avoided making behavioral changes to the program, but because all the code was being moved around, I noticed a bunch of quirks with the help system that I've fixed along the way (because the entire help system was rewritten during the refactor). Behavioral changes as a result are:

* Help has been added to commands with missing help. Missing flags have also been added to some functions.
* `shield flags` has been destroyed. It used to print out the flags list, as default by the parser. It was the _old old old old_ behavior for help, and it was mostly kept in just to have it around. Its not useful, and so its now gone in favor of listing the flags for each function in help
* Giving an improper flag no longer spams out the old `shield flags` output. Instead, it tells the user simply that the flag they specified wasn't recognized.
* `shield commands` output is now at `shield help`, because its probably what you wanted when you asked for help anyway. `shield commands` is now just an alias for `shield help`.
* The `-h` or `--help` flags now work like you would expect them to in any program. `shield -h` is the same as `shield help`, and `shield <command> -h` is the same as `shield help <command>`. It actually doesn't matter in what order you put the flag, its the same as putting `help` at the front no matter what.
* If there's no backend specified, and the user gave a nonsense command, the program would choose to complain about the missing backend. It now favors complaining about the unrecognized command.


thomasmmitchell:
Yes, the build is failing. Godep is giving me a headache. Just a moment.

PR #317, opened 2017-09-19
Command parser chooses longest possible command

thomasmmitchell:
`shield restore archive 01234567-89ab-cdef-0123-456789abcdef` should work, not complain that archive isn't a uuid because `restore` another name for that command. It's just better this way.


thomasmmitchell:
Still in Godep hell, I guess. Will fix.

PR #320, opened 2017-09-25
Bbr plugin v2

ramonskie:
a new try to get bbr plugin integrated with shield


jhunt:
Looks like the Travis checks are failing...

ramonskie:
yeah i was working on it.
its because of missing dependencies.

the vendor directory is getting pretty big so i'm not sure if i should change it to using the bbr cli instead of the go code. @jhunt your opinion?

jhunt:
Given the scope being introduced here, along with other parallel efforts to reduce the number of external library deps, I would welcome a change to use the bbr CLI.

Benefits:
1. We can package it with the BOSH release without issue
2. The plugin framework already has fairly nice facilities for executing other programs (look at `fs`)
3. We (SHIELD) are then using BBR the same way operators are, in case they need/want to interact with stuff out-of-band

:+1:

PR #324, opened 2017-09-27
Reorganized Vault code

daviddob:
Pulled vault out of core and into crypter, provided better error messages when unsealing/init vault from cli, made passing vault creds expandable


crypter/vault.go
```diff
@@ -31,10 +33,34 @@ type VaultCreds struct {
...
+			},
+		},
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if len(via) > 10 {
+				return fmt.Errorf("stopped after 10 redirects")
+			}
+			req.Header.Add("X-Vault-Token", new_vault.Token)
+			return nil
+		},
+	}
```
jhunt: You can inline the `http.Client{}` intializer

crypter/vault.go
```diff
@@ -31,10 +33,34 @@ type VaultCreds struct {
...
+
+	return new_vault, nil
```
jhunt: You can also just return the literal, at that point.

crypter/vault.go
```diff
@@ -31,10 +33,34 @@ type VaultCreds struct {
 	RootToken string `json:"root_token"`
 }
 
-var status struct {
+var Status struct {
```
jhunt: Why is this exported?

crypter/vault.go
```diff
@@ -17,6 +18,7 @@ import (
 	"strings"
 
 	"github.com/starkandwayne/goutils/log"
+	"golang.org/x/net/lex/httplex"
```
jhunt: What is httplex?

daviddob: httplex is the library used to determine if http header names/vaules are valid. Im using it to determine if the decryption of the vault creds occurred properly and avoid sending vault invalid headers (it was causing errors where the daemon had to be reset as the vault token header would be invalid and all calls to vault thereafter would fail)

crypter/vault.go
daviddob: good question, I exported the status function and apparently got a bit export happy

crypter/vault.go
jhunt: Is this because an invalid master password leads to garbage binary data?  If so, I'd recommend a two-pronged approach:

1. Store a "sigil" value in the plaintext before decryption -- something known to SHIELD that is not used for anything other than "did we get a good decrypt".  Bad master passwords will not decrypt properly, so this known value will not match what's in the software.
2. Use a regex to validate the master password to ensure that it only contains printable chars (ASCII space (0x20) to ASCII '~' (0x7e), something like `/^[\x20-\x7e]+$/`

Issue #278, opened 2017-05-05
Allow "shield create backend" with --ca-cert


drnic:
In https://github.com/starkandwayne/shield-boshrelease/tree/links-and-cloud-config it will be easy/default to deploy SHIELD with TLS and a self-signed certificate. 

It would be delicious for `shield create-backend` to support a `--ca-cert` flag to consume the root ca.

thomasmmitchell:
@drnic I'm looking to potentially target this soon. Can you somehow recall what you were talking about several months ago and give more information / an example on expected usage?

drnic:
@thomasmmitchell in our profession we would help everyone if we moved ppl away from `-k` or any variation of "skip ssl/tls validation" in our clients.

For example, https://github.com/cloudfoundry/bosh-cli provides `bosh --ca-cert path/to/cert alias-env` to cache the root CA for all future client interactions. `bosh2` no longer supports `-k` or any SSL validation skipping.

The CF CLI team is also starting to investigate adding support for specifying custom root CAs.

Can we move towards this too and eventually get rid of `-k`?

thomasmmitchell:
I've never used `--ca-cert` on bosh2, but if I'm imagining a concept that solves this problem, I'm assuming that you give it that root CA to treat as trusted, such that your self-signed cert on the deployment is accepted, while someone else's garbage would shoot back with a TLS issue. That's a good idea, and I would definitely like that feature.

As for getting rid of `-k`... it's not happening _soon_. However, once `--ca-cert` is a thing, I'm fine with calling it deprecated... or possibly putting a quick "you should feel bad, use this other thing" line when folks use `-k` or a backend marked to skip TLS verification. I'm not married to the idea of the `-k` flag, even though I have to use it more often than I'd like. In the end, I'm not sure if I'll remove the flag entirely, so much as I might just stop advertising it. I'll have to see if a `-k`less workflow causes issues in any contexts. Either way, the final decision on its removal is one for another day... say... pending major version 9.

jhunt:
This has been merged into v8

drnic:
@thomasmmitchell how do you use `bosh2` without `bosh alias-env --ca-cert` or setting `$BOSH_CA_CERT`?

drnic:
thanks for adding --ca-cert

thomasmmitchell:
I either use it or have an environment where the cert is legitimate

PR #325, opened 2017-09-28
Cli ca cert

thomasmmitchell:
#278 
--ca-cert is now implemented. It specifies a root CA to trust. If given during `create-backend`, that backend will always use it.
If used otherwise, it will override whatever the backend setting is.

The flag takes a value. If the value is a PEM encoded cert, it will use it directly. If its not, it will be interpreted as a file path which should contain a PEM encoded cert therein.

Also, in order to see things about the configured certs, `backends --full` has been implemented. It shows more verbose info about the configured backends, including the auth token and the ca-cert. The ca-cert column shows the plaintext subject, SANs, and validity time range.

### DIFF
--- a/cmd/shield/commands/backends/list.go
+++ b/cmd/shield/commands/backends/list.go
@@ -1,10 +1,12 @@
 package backends
 
 import (
+	"fmt"
 	"os"
 	"sort"
 
 	"github.com/starkandwayne/goutils/ansi"
+	"github.com/starkandwayne/shield/api"
 	"github.com/starkandwayne/shield/cmd/shield/commands"
 	"github.com/starkandwayne/shield/cmd/shield/commands/internal"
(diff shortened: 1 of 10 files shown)


cmd/shield/commands/backends/list.go
```diff
@@ -46,13 +76,116 @@ func cliListBackends(opts *commands.Options, args ...string) error {
...
+	return &t, nil
+}
+
+func greenify(s string) string {
+	return ansi.Sprintf("@G{%s}", s)
+}
+
+func redify(s string) string {
+	return ansi.Sprintf("@R{%s}", s)
+}
```
jhunt: No.  Just no.

Remove the greenify and redify functions, and use the inline calls to `ansi.Sprintf`

PR #327, opened 2017-09-28
Added seal status checks to individual api endpoints

daviddob:
Added assertions into the core module for vault sealed status, returns 403 if sealed

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -17,8 +17,8 @@ import (
 	"github.com/starkandwayne/goutils/log"
 
 	"github.com/starkandwayne/shield/db"
-	"github.com/starkandwayne/shield/util"
 	"github.com/starkandwayne/shield/timespec"
+	"github.com/starkandwayne/shield/util"
 )
 
 //APIVersion is the maximum supported version of the core Shield Daemon API.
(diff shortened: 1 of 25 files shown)


core/api.go
```diff
@@ -64,163 +66,301 @@ func (core *Core) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-	}
-
-	if sealed, _ := core.vault.IsSealed(); sealed == false {
-		switch {
-		case match(req, `GET /v1/meta/pubkey`):
-			core.v1GetPublicKey(w, req)
+	//All api endpoints below have the mustBeUnlocked requirement such that if vault
+	//	is sealed or uninitialized they will return a 403
+	case match(req, `GET /v1/meta/pubkey`):
+		if core.mustBeUnlocked(w) {
```
jhunt: What if we try:

```
if locked := core.mustBeUnlocked(); locked {
    // ...
}
```

core/api.go
```diff
@@ -303,6 +443,21 @@ func invalidlimit(limit string) bool {
 	return false
 }
 
+func (core *Core) mustBeUnlocked(w http.ResponseWriter) bool {
+	status, err := core.vault.status()
+	if err != nil {
+		bail(w, err)
```
jhunt: You need a `return true` here.

jhunt: Went ahead and took care of this one.

jhunt:
I merged this out-of-band and squash/rebased it onto `v8`.

PR #329, opened 2017-10-02
Encryption Key Rotation

daviddob:
Implementation in place for key rotation via API, CLI or UI
Could probably use more CSS styling before final release

### DIFF
--- a/cmd/shield/commands/access/rotate.go
+++ b/cmd/shield/commands/access/rotate.go
@@ -11,35 +11,37 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-//Init - Initializes the encryption key database
-var Init = &commands.Command{
-	Summary: "Initialize the encryption key database",
+//Rotate - Rotates the encryption database keys
+var Rotate = &commands.Command{
+	Summary: "Rotate the encryption database keys",
 	Help:    &commands.HelpInfo{},
-	RunFn:   cliInit,
+	RunFn:   cliRotate,
 	Group:   commands.AccessGroup,
 }
 
-func cliInit(opts *commands.Options, args ...string) error {
-	log.DEBUG("running 'init' command")
+func cliRotate(opts *commands.Options, args ...string) error {
+	log.DEBUG("running 'rotate' command")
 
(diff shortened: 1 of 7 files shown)


cmd/shield/commands/access/rotate.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/starkandwayne/goutils/ansi"
+	"github.com/starkandwayne/shield/api"
+	"github.com/starkandwayne/shield/cmd/shield/commands"
+	"github.com/starkandwayne/shield/cmd/shield/commands/internal"
+	"github.com/starkandwayne/shield/cmd/shield/log"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+//Rotate - Rotates the encryption database keys
```
jhunt: Can we call this `shield rekey-master`?  @thomasmmitchell what do you think?

core/api.go
```diff
@@ -444,7 +448,7 @@ func invalidlimit(limit string) bool {
 }
 
 func (core *Core) mustBeUnlocked(w http.ResponseWriter) bool {
-	status, err := core.vault.status()
+	status, err := core.vault.Status()
```
jhunt: ... curses, I fixed this in my branch already.  I'll deal with the merge conflicts..

cmd/shield/commands/access/rotate.go
daviddob: Its a bit more intuitive than `shield rotate` imo

thomasmmitchell: rekey is more consistent with `safe`, and I think rotate alone is a little too vague. SHIELD isn't for dealing with keys, so rotate alone is like... rotate what. `shield rekey` might be fine. `rekey-master` may be more verbose than needed, unless we're going to be rekeying other things which doesn't... make sense (?).

jhunt: I figured `rekey-master` was better because it's longer and more annoying to type.  My first inclination was `shield change-the-master-password-yes-that-one-the-super-important-one`

PR #332, opened 2017-10-03
Rework v2 API with new `route' framework

jhunt:
`route` is a lightweight framework for making it easier and less
error-prone to write and implement.  All of the previously implemented
`/v2/*` namespace has been re-written in the new style.


jhunt:
BTW, Travis Tests _should_ pass on this one...

daviddob:
Looks good to me.

PR #336, opened 2017-10-04
New `shield curl` command

jhunt:
This new command allows operators and SHIELD hackers to interact
directly with the SHIELD API, without interpreting the JSON result.
Responses from the SHIELD API will be printed, with some formatting.

Example:

    $ shield curl /v2/health
    $ shield curl POST /v2/tenants/create <tenant.json

### DIFF
--- a/cmd/shield/commands/misc/curl.go
+++ b/cmd/shield/commands/misc/curl.go
@@ -1,0 +1,81 @@
+package misc
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/mattn/go-isatty"
+
+	"github.com/starkandwayne/shield/api"
+	"github.com/starkandwayne/shield/cmd/shield/commands"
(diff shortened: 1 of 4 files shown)


cmd/shield/commands/misc/curl.go
```diff
@@ -0,0 +1,76 @@
...
+			method = args[0]
+			url = args[1]
+
+		} else {
+			return fmt.Errorf("USAGE: shield curl [METHOD] URL\n")
+		}
+
+		if method == "PUT" || method == "PATCH" || method == "POST" {
+			commands.MSG("Please enter your JSON (^D when done)")
+			body, err = internal.ReadAll(os.Stdin)
```
thomasmmitchell: Nitpick inc... So, instead of stalling on stdin, we could do this more curl-esque and check if stdin is a terminal first, and only if it isn't (assume its a pipe / redirect), then we read it. I'm not sure if you'd want to interactively write JSON but... up for debate.

PR #337, opened 2017-10-04
Drop Support for MySQL and PostgreSQL

jhunt:
SHIELD now only supports SQLite3


drnic:
:trollface: how do I run a 3-node cluster of sqlite3?

jhunt:
This moves us into a position to introduce a new `shield` backup plugin in v8, to back up the Vault encryption store and the SQLite3 database file in one go.

PR #340, opened 2017-10-06
Localauth API and Web UI changes

thomasmmitchell:
In preparation for CLI support

### DIFF
--- a/core/auth.go
+++ b/core/auth.go
@@ -56,25 +56,34 @@ type authResponse struct {
 	Tenants []authTenant `json:"tenants"`
 }
 
-func (core *Core) checkAuth(req *http.Request) (*authResponse, error) {
-	cookie, err := req.Cookie(SessionCookieName)
-	if err != nil {
-		if err == http.ErrNoCookie {
-			log.Debugf("no session cookie ('%s') found in request; treating as unauthenticated", SessionCookieName)
-		} else {
-			log.Debugf("failed to extract session cookie ('%s') from request: %s", SessionCookieName, err)
+//Gets the session ID from the request. Returns "" if not given.
+func getSessionID(req *http.Request) string {
+	sessionID := req.Header.Get("X-Shield-Session")
+
+	if sessionID == "" { //If not in header, check for cookie
+		cookie, err := req.Cookie(SessionCookieName)
+		if err == nil { //ErrNoCookie is the only error returned from Cookie()
+			sessionID = cookie.Value
 		}
+	}
+	return sessionID
+}
+
+func (core *Core) checkAuth(req *http.Request) (*authResponse, error) {
+	sessionID := getSessionID(req)
+	if sessionID == "" {
+		log.Debugf("No session ID was found in request")
 		return nil, nil
 	}
 
-	log.Debugf("retrieving user account for session '%s'", cookie.Value)
-	user, err := core.DB.GetUserForSession(cookie.Value)
+	log.Debugf("retrieving user account for session '%s'", sessionID)
+	user, err := core.DB.GetUserForSession(sessionID)
 	if err != nil {
-		log.Debugf("failed to retrieve user account for session '%s': %s", cookie.Value, err)
+		log.Debugf("failed to retrieve user account for session '%s': %s", sessionID, err)
 		return nil, err
 	}
 	if user == nil {
-		log.Debugf("failed to retrieve user account for session '%s': database did not throw an error, but returned a nil user", cookie.Value)
+		log.Debugf("failed to retrieve user account for session '%s': database did not throw an error, but returned a nil user", sessionID)
 		return nil, nil
 	}
 
--- a/core/auth.go
+++ b/core/auth.go
@@ -69,13 +69,7 @@ func getSessionID(req *http.Request) string {
 	return sessionID
 }
 
-func (core *Core) checkAuth(req *http.Request) (*authResponse, error) {
-	sessionID := getSessionID(req)
-	if sessionID == "" {
-		log.Debugf("No session ID was found in request")
-		return nil, nil
-	}
-
+func (core *Core) checkAuth(sessionID string) (*authResponse, error) {
 	log.Debugf("retrieving user account for session '%s'", sessionID)
 	user, err := core.DB.GetUserForSession(sessionID)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


core/auth.go
```diff
@@ -56,25 +56,34 @@ type authResponse struct {
...
-			log.Debugf("no session cookie ('%s') found in request; treating as unauthenticated", SessionCookieName)
-		} else {
-			log.Debugf("failed to extract session cookie ('%s') from request: %s", SessionCookieName, err)
+//Gets the session ID from the request. Returns "" if not given.
+func getSessionID(req *http.Request) string {
+	sessionID := req.Header.Get("X-Shield-Session")
+
+	if sessionID == "" { //If not in header, check for cookie
+		cookie, err := req.Cookie(SessionCookieName)
+		if err == nil { //ErrNoCookie is the only error returned from Cookie()
```
jhunt: I would believe that now, but in the future?

core/v2.go
```diff
@@ -698,6 +699,84 @@ func (core *Core) v2API() *route.Router {
...
+			Username string
+			Password string
+		}{}
+
+		if !r.Payload(&auth) { //Payload reports its own errors
+			return
+		}
+
+		if auth.Username == "" {
+			r.Fail(route.Errorf(http.StatusForbidden, nil, "no username given"))
```
jhunt: Use `403` here

core/v2.go
```diff
@@ -698,6 +699,84 @@ func (core *Core) v2API() *route.Router {
...
+			return
+		}
+
+		if auth.Password == "" {
+			r.Fail(route.Errorf(http.StatusForbidden, nil, "no password given"))
```
jhunt: Use `403` here

core/v2.go
```diff
@@ -698,6 +699,84 @@ func (core *Core) v2API() *route.Router {
...
+		}
+
+		user, err := core.DB.GetUser(auth.Username, "local")
+		if err != nil {
+			r.Fail(route.Oops(err, "An unknown error occurred when authenticating local user '%s'", auth.Username))
+			return
+		}
+
+		if user == nil || !user.Authenticate(auth.Password) {
+			r.Fail(route.Errorf(http.StatusForbidden, nil, "Incorrect username or password"))
```
jhunt: also here

core/v2.go
```diff
@@ -698,6 +699,84 @@ func (core *Core) v2API() *route.Router {
...
+		r.SetCookie(SessionCookie(session.UUID.String(), true))
+		r.SetHeader("X-Shield-Session", session.UUID.String())
+
+		r.Success("Logged in as user %s", auth.Username)
+	})
+
+	r.Dispatch("GET /v2/auth/logout", func(r *route.Request) {
+		sessionID := getSessionID(r.Req)
+		if sessionID == "" {
+			//I guess we're okay with not getting a session to logout?...
```
jhunt: Can we be a little more assertive here?

core/v2.go
```diff
@@ -698,6 +699,84 @@ func (core *Core) v2API() *route.Router {
...
+			r.Success("No user to logout")
+		}
+
+		id := uuid.Parse(sessionID)
+		if id == nil {
+			r.Fail(route.Bad(nil, "Invalid session ID received"))
+		}
+		err := core.DB.ClearSession(id)
+		if err != nil {
+			r.Fail(route.Oops(err, "An unknown error occurred when logging out of the session"))
```
jhunt: Drop everything after "occurred"

core/v2.go
```diff
@@ -698,6 +699,84 @@ func (core *Core) v2API() *route.Router {
...
+	r.Dispatch("GET /v2/auth/logout", func(r *route.Request) {
+		sessionID := getSessionID(r.Req)
+		if sessionID == "" {
+			//I guess we're okay with not getting a session to logout?...
+			r.Success("No user to logout")
+		}
+
+		id := uuid.Parse(sessionID)
+		if id == nil {
+			r.Fail(route.Bad(nil, "Invalid session ID received"))
```
jhunt: We should probably formulate an `error` object to pass to `Bad()`, and make the FE error message more generic, like "Unable to log out"

web2/htdocs/index.html
```diff
@@ -1433,10 +1435,24 @@
         (function () {
           api({
             type: 'GET',
-            url:  '/v2/auth/providers',
+            url: '/v2/auth/providers',
```
jhunt: Fix the formatting regression here.

web2/htdocs/index.html
```diff
@@ -1433,10 +1435,24 @@
...
             success: function (data) {
-              $('#viewport').html(template('login', { oauth: data }));
-            },
+              $('#viewport').html(template('login', { oauth: data }))
+                .on("submit", "form", function (event) {
+                event.preventDefault()
+                api({
+                  url: "/v2/auth/login",
+                  method: "POST",
```
jhunt: This should be `type` not `method`, and it should be listed first.

web2/htdocs/index.html
```diff
@@ -1433,10 +1435,24 @@
...
+                  data: $(event.target).serializeObject(),
+                  success: function () {
+                    document.location.href = "/";
```
jhunt: Make this a `goto("")` instead of messing with location directly

web2/htdocs/index.html
```diff
@@ -1445,6 +1461,25 @@
         break; /* #!/login */
         // }}}
 
+      case "#!/logout": // {{{
+        (function () {
+          api({
+            url: "/v2/auth/logout",
```
jhunt: You need a `type` here, of `GET`

web2/htdocs/index.html
```diff
@@ -1445,6 +1461,25 @@
...
+            success: function () {
+              document.location.href = "/";
```
jhunt: Use `goto("")` instead of changing document.location directly

web2/htdocs/index.html
```diff
@@ -1445,6 +1461,25 @@
...
+            },
+            error: function (xhr) {
+              if (xhr.status >= 500) {
+                $('#viewport').html(template('BOOM'));
+              } else {
+                document.location.href = "/";
```
jhunt: Use `goto("")`

core/auth.go
thomasmmitchell: "Cookie returns the named cookie provided in the request or ErrNoCookie if not found. "

web2/htdocs/index.html
thomasmmitchell: Are we changing the behavior of api in the future? GET is the documented default method of .ajax

jhunt: It's a good idea to be explicit when it aids understanding.

PR #341, opened 2017-10-09
CLI Local User Management

daviddob:
Added CLI and Testing Support for Local User Management

New commands added:
`shield create-user`
`shield edit-user`
`shield delete-user`
`shield users`
`shield user <uuid>`


cmd/shield/commands/dispatcher.go
```diff
@@ -25,16 +25,18 @@ type HelpGroup struct {
...
-	JobsGroup     = &HelpGroup{name: "JOBS"}
-	ArchivesGroup = &HelpGroup{name: "ARCHIVES"}
-	AccessGroup   = &HelpGroup{name: "ACCESS"}
-	TasksGroup    = &HelpGroup{name: "TASKS"}
+	InfoGroup      = &HelpGroup{name: "INFO"}
+	MiscGroup      = &HelpGroup{name: "MISCELLANEOUS"}
+	BackendsGroup  = &HelpGroup{name: "BACKENDS"}
+	TargetsGroup   = &HelpGroup{name: "TARGETS"}
+	StoresGroup    = &HelpGroup{name: "STORES"}
+	SchedulesGroup = &HelpGroup{name: "SCHEDULES"}
```
jhunt: You should not be re-adding the SchedulesGroup -- maybe need a rebase against v8?

cmd/shield/main.go
```diff
@@ -52,10 +53,14 @@ func main() {
 		After:     getopt.StringLong("after", 'A', "", "Only show archives that were taken after the given date, in YYYYMMDD format."),
 		Before:    getopt.StringLong("before", 'B', "", "Only show archives that were taken before the given date, in YYYYMMDD format."),
 		To:        getopt.StringLong("to", 0, "", "Restore the archive in question to a different target, specified by UUID"),
-		Limit:     getopt.StringLong("limit", 0, "", "Display only the X most recent tasks or archives"),
+		Limit:     getopt.StringLong("limit", 0, "", "Display only the X most recent tasks, archives, or users"),
 
 		Full: getopt.BoolLong("full", 0, "Show all backend information when listing backends"),
 
+		Backend: getopt.StringLong("backend", 'b', "", "Only show users with the specified backend."),
```
jhunt: Where is this option used?  Looks to me like `shield users` only shows local users...

t/api
```diff
@@ -502,8 +503,57 @@ try "Checking that various API endpoints require JSON payloads" # {{{
 
 # }}}
 
+# }}}
+#
+try "Checking that unimplemented v2HTTP methods return 501s" # {{{
```
jhunt: missing space between `v2` and `HTTP`

t/api
```diff
@@ -502,8 +503,57 @@ try "Checking that various API endpoints require JSON payloads" # {{{
...
+   for type in auth/local/users; do
+     is "$(httpstat $method "http://$SHIELD_DAEMON_ADDR/v1/$type/sub/requests")" \
+        "501" "/v2/$type/sub/requests should HTTP 501 (not implemented) on a $method request"
+   done
+ done
+) 2>&1 | indent
+
+
+# }}}
+try "Checking that various v2API endpoints validate UUIDs" # {{{
```
jhunt: missing space between `v2` and `API`

t/api
```diff
@@ -502,8 +503,57 @@ try "Checking that various API endpoints require JSON payloads" # {{{
...
+       is "$(httpstat $method "http://$SHIELD_DAEMON_ADDR/v1/$type/$uuid")" \
+          "501" "$method /v2/$type/$uuid should HTTP 501 (not implemented) on a malformed UUID"
+     done
+   done
+ done
+) 2>&1 | indent
+
+
+# }}}
+try "Checking that various v2API endpoints require JSON payloads" # {{{
```
jhunt: missing space between `v2` and `API`

t/api
```diff
@@ -636,6 +686,26 @@ EOF
   "paused"    : false
 }
 EOF
+
+ create user exampleuser <<EOF
+{
+  "uuid":"355ccd3f-1d2f-49d5-937b-f4a12033a0cf",
+	"name":"exampleuser",
```
jhunt: spacing issue

t/api
```diff
@@ -636,6 +686,26 @@ EOF
...
+  "account":"exampleuser",
+  "password":"foobar",
+  "sysrole":"technician"
+}
+EOF
+
+ create user admin <<EOF
+{
+  "uuid":"355ccd3f-1d2f-49d5-937b-f4a12033a0ce",
+	"name":"admin",
```
jhunt: spacing issue

t/api
```diff
@@ -949,6 +1019,64 @@ try "Pausing and unpausing jobs" # {{{
...
+
+ run users --fuzzy a
+ context 'users with account "*a*"'
+ includes user exampleuser
+ includes user admin
+
+ run users --fuzzy m
+ context 'users with account "*m*"'
+ includes user exampleuser
+ includes user admin
```
jhunt: i would try `ex` instead of `m` and verify that we exclude admin, but include exampleuser

t/api
```diff
@@ -949,6 +1019,64 @@ try "Pausing and unpausing jobs" # {{{
...
+
+ run users --sysrole technician
+ context 'users with sysrole technician'
+ includes user exampleuser
+ excludes user admin
+
+ run users --sysrole manager
+ context 'users with sysrole manager'
+ none 'no users with sysrole "manager"'
+
```
jhunt: We have no --backend tests -- is `--backend` being honored?

t/api
```diff
@@ -1025,6 +1153,24 @@ EOF
   "paused"    : false
 }
 EOF
+
+update user admin <<EOF
+{
+  "uuid":"355ccd3f-1d2f-49d5-937b-f4a12033a0ce",
+	"name":"nobody",
```
jhunt: spacing issue

t/api
```diff
@@ -1045,6 +1191,11 @@ try "Deleting things" # {{{
  can_delete target   redis-shared
  can_delete store    fs
  can_delete policy   short-term
+
+ context 'users can always be deleted'
+ can_delete user admin
+ can_delete user exampleuser
```
jhunt: Can we delete a non-existent user?

cmd/shield/main.go
daviddob: Yeah thats true, I had functionality to list users from all backends but hardcoded it to local as the request was for LUM, (not sure if we want to include that via a backend flag to show other than local)

t/api
daviddob: see above comment

cmd/shield/main.go
daviddob: You also (shouldnt/wouldnt) be able to edit/delete/etc non local backend users as that is all handled via the mapping.

PR #344, opened 2017-10-11
Token Authentication Provider

jhunt:
This commit introduces a new authentication provider for doing Token
Authentication via the CLI.  It also introduces a split in how auth
providers are listed, so that admins can configure certain auth
providers to only work via the Web UI or the CLI.

### DIFF
--- a/bin/testdev
+++ b/bin/testdev
@@ -123,9 +123,21 @@ slow_loop:     20
 fast_loop:     2
 vault_keyfile: ${workdir}/var/vault.crypt
 auth:
+  - name:       Token
+    identifier: token
+    backend:    token
+    usage:      cli
+    properties:
+      4E7E8965-9DB0-4D3A-B920-2537F3321B3D:
+        - tenant: starkandwayne
+          role:   engineer
+        - tenant: CF Commuity
+          role:   engineer
+
   - name:       Github
     identifier: public-gh
     backend:    github
+    usage:      web
     properties:
       client_id:     7760ef75ba49c9699114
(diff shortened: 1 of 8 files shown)


jhunt:
In thinking more about this "two configs for the price of one" setup, the weirdness just comes out the other end, because now I have to register two OAuth applications (one for web / one for cli) to make things work.  Either that, or we do the "mirroring" when we read the config, but that leads to confusion as to why the doubling up.

Thoughts?

thomasmmitchell:
Sure, you have to register two OAuth apps, but you already need to do that. I guess the problem here is just the messaging to the user as to why they need to register twice? The alternative is changing the  way that we handle OAuth such that it can work with both UIs from one logical client.

jhunt:
After much discussion, I think we have a way forward that doesn't involve this PR:

- Modify the Provider Initiator URLs to be able to differentiate between CLI / Web users (via `?web`)
- Have the Initiators set a cookie indicating which mode the user entered the flow in (web or cli)
- Read (and unset) the cookie on the Provider Redirect URLs, and react appropriately:
   - Web Users will still get redirected with a Set-Cookie header
   - CLI Users will see a prett-ified, SHIELD-branded page that shows them the session ID for copying / pasting into the waiting terminal (or we use the local port hack)
- This Token Provider will continue to be CLI only, but that disposition will be _in code_, not in configuration.

This approach requires neither split configuration, nor separate OAuth2 application registrations with the 3rd party systems, so it's a win-win!

PR #345, opened 2017-10-11
CLI added login, logout, and whoami

thomasmmitchell:
* Auth prompting was taken out of the api package and moved into CLI
commands. A user now needs to authenticate manually with one of these
commands, as opposed to the old way, in which the user would simply be
prompted upon receiving a 401
* API functions were added for contacting POST /v2/auth/login, GET
/v2/auth/logout, and GET /v2/auth/id.
* Parsing functions for getting v1 and v2 API error responses were
added. v1 endpoints on v2 APIs will currently be unable to get error
messages, but this will be okay once the v1 endpoints go away.
* Error types were added for 401s and 403s, namely ErrUnauthorized and
ErrForbidden. These can be type checked to determine if an api error was
due to authentication problems.
* The core API exposes test endpoints if the debug flag in the daemon
config is set.
* The v2 api now gives 401s if the user is not authenticated (as opposed
to 403)
* The api.Backend type now accepts an APIVersion int, for being told
which APIVersion to use. This value can be stored in the in-memory
version of the config for use in the rest of the CLI, but is never written out
to disk.
* testdev now sets the debug flag for the shield daemon config.
* The login endpoint now gives back information about the user which was
authenticated in the same format as /auth/id
* api_version is no longer exposed by the status endpoint (because you
need to be authenticated to reach it), and is instead exposed by
/v1/ping

### DIFF
--- a/api/access.go
+++ b/api/access.go
@@ -133,7 +133,7 @@ func Login(username, password string) (sessionID string, user *AuthIDOutput, err
 		user = &AuthIDOutput{}
 		err = json.Unmarshal(body, user)
 	} else {
-		err = getV2Error(r)
+		err = getAPIError(r)
 	}
 
 	return
@@ -170,7 +170,7 @@ func LogoutSession(sessionID string) error {
 	}
 
 	if r.StatusCode != 200 {
-		err = getV2Error(r)
+		err = getAPIError(r)
 	}
 
 	return err
--- a/api/access.go
+++ b/api/access.go
@@ -3,7 +3,6 @@ package api
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"io/ioutil"
 	"net/http"
 
@@ -27,9 +26,6 @@ func Unlock(master string) error {
 	}
 	respMap := make(map[string]string)
 	if err := uri.Post(&respMap, string(contentJSON)); err != nil {
-		if initError, present := respMap["error"]; present {
-			return errors.New(initError)
-		}
 		return err
 	}
 
@@ -53,9 +49,6 @@ func Init(master string) error {
 		return err
 	}
 	if err := uri.Post(&respMap, string(contentJSON)); err != nil {
-		if initError, present := respMap["error"]; present {
-			return errors.New(initError)
-		}
 		return err
 	}
 
@@ -81,9 +74,6 @@ func Rekey(current, proposed string) error {
 		return err
 	}
 	if err := uri.Post(&respMap, string(b)); err != nil {
-		if rekeyError, present := respMap["error"]; present {
-			return errors.New(rekeyError)
-		}
 		return err
 	}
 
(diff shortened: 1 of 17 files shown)


api/access.go
```diff
@@ -22,8 +27,8 @@ func Unlock(master string) error {
 	}
 	respMap := make(map[string]string)
 	if err := uri.Post(&respMap, string(contentJSON)); err != nil {
-		if init_error, present := respMap["error"]; present {
-			return errors.New(init_error)
+		if initError, present := respMap["error"]; present {
+			return errors.New(initError)
```
jhunt: Can you remove these superfluous changes?

api/access.go
```diff
@@ -76,11 +81,131 @@ func Rekey(current, proposed string) error {
...
+	j, err := json.Marshal(struct {
+		Username string `json:"username"`
+		Password string `json:"password"`
+	}{
+		Username: username,
+		Password: password,
+	})
+
+	if err != nil {
+		panic("Could not marshal auth struct that we JUST made")
```
jhunt: Should we return the error instead of panic()-ing?

api/access.go
```diff
@@ -76,11 +81,131 @@ func Rekey(current, proposed string) error {
...
+		err = getAPIError(r)
+	}
+
+	return err
+}
+
+//AuthIDOutput contains all the information from a call to /v2/auth/id
+type AuthIDOutput struct {
+	User    AuthUser     `json:"user"`
+	Tenants []AuthTenant `json:"tenants"`
```
jhunt: You ought to be able to nest these and avoid creating visible structs that are useless outside of the context of an AuthID response...

api/status.go
```diff
@@ -1,9 +1,14 @@
...
+	"net/http"
+	"strings"
+)
+
 type Status struct {
-	Name       string `json:"name"`
-	Version    string `json:"version"`
-	APIVersion int    `json:"api_version"`
+	Name    string `json:"name"`
+	Version string `json:"version"`
```
jhunt: Can we just make the status API not auth-protected?

api/access.go
thomasmmitchell: No. On a json Marshal, the only way that err should occur is if the pointer is nil or if the argument isn't a pointer. Both of those are completely in our control, and this is an assertion that we didn't botch that.

api/status.go
thomasmmitchell: We sure could. Seemed like you didn't want to before, though. Just wanna do it anyway now?

api/access.go
thomasmmitchell: Hey. This is golang we're talking about. What's today simply caught by `golint` is tomorrow mandated by the compiler.

Fine...

thomasmmitchell: Actually, I realized that entire if is unnecessary because I'm handling the error parsing in Request now

api/status.go
jhunt: Yeah.  We just need to remove the Version - set an omitempty, and dont set it if we aren't authenticated.

thomasmmitchell: I'll put APIVersion back here. The auth hiding isn't in scope here because... v2 doesn't know what auth is yet

jhunt: Right.  Unauthenticated sessions should receive:

```json
{
  "name"        : "My Awesome SHIELD",
  "api_version" : 2
}
```

Authenticated sessions should receive:

```json
{
  "name"        : "My Awesome SHIELD",
  "api_version" : 2,
  "version"     : "8.0.3"
}
```

thomasmmitchell: You know... I'll put it in here. We'll leave status unauthed and I'll do manual auth inside so we can have the special handling.

thomasmmitchell:
@jhunt resume the reviewing

thomasmmitchell:
@jhunt *waves hands* it was rebased _all along_

jhunt:
Oh Wow, I Don't Know What Pull Request I Was Looking At ™️

PR #347, opened 2017-10-12
Combine Web and CLI Authentication Flows

jhunt:
We can now support both CLI and Web Auth Provider flows.
Each authentication provider now has two entry point URIs,
once for web users and another for CLI users (by way of a
web browser, to make it all work).  As people enter one way
or another, they get a `via` cookie that tracks their particular
entry point, so we can handle them on redirect properly.

On redirect, web users continue to get a `Set-Cookie` header
in their final redirect back to SHIELD's `/` page.  CLI users
will instead get a new page, `#!/cliauth` that presents them
with their newly-minted session ID, and instructs them to
copy / paste it into their SHIELD CLI (none of the receiving
end work has been done yet, an is out of scope for this PR).


core/github_auth.go
```diff
@@ -29,130 +30,111 @@ type GithubAuthProvider struct {
 		} `json:"rights"`
 	} `json:"mapping"`
 
-	Name       string
-	Identifier string
-	Usage      string
-	core       *Core
+	Name  string
+	Usage string
```
thomasmmitchell: What are we still using Usage for?

core/github_auth.go
```diff
@@ -29,130 +30,111 @@ type GithubAuthProvider struct {
...
+	core  *Core
 }
 
-func (gh *GithubAuthProvider) DisplayName() string {
-	return gh.Name
-}
+func (p *GithubAuthProvider) Configure(raw map[interface{}]interface{}) error {
+	p.Type = "github"
```
thomasmmitchell: Is this redundant? This is set in auth.go with the rest of the AuthProviderBase

thomasmmitchell:
How are we determining that Token shouldn't be shown in the web ui?

core/api.go
```diff
@@ -28,8 +28,55 @@ func (core *Core) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			w.WriteHeader(404)
+			fmt.Fprintf(w, "couldn't find that auth provider...: %s", err)
+			return
+		}
+
+		if m[2] == "redir" {
+			via := "web"
+			if cookie, err := req.Cookie("via"); err == nil {
+				via = cookie.Value
+			}
```
thomasmmitchell: Should we kill the via Cookie at this point

core/v2.go
```diff
@@ -225,28 +221,34 @@ func (core *Core) v2API() *route.Router {
...
+		typ := r.Param("for", "cli")
 		for _, auth := range core.auth {
-			if usage == "" || auth.Usage == usage {
-				l = append(l, v2AuthProvider{
-					Name:       auth.Name,
-					Identifier: auth.Identifier,
-					Type:       auth.Backend,
-					Usage:      auth.Usage,
-				})
+			if auth.Backend == "token" && typ != "cli" {
```
jhunt: @thomasmmitchell this is where we hide the token backend from the list of available auth providers, unless you want CLI providers.

core/api.go
```diff
@@ -28,8 +28,55 @@ func (core *Core) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+				fmt.Fprintf(w, "The authentication process broke down\n")
+				w.WriteHeader(500)
+			}
+
+			session, err := core.createSession(user)
+			if err != nil {
+				log.Errorf("failed to create a session for user %s@%s: %s", user.Account, user.Backend, err)
+				w.Header().Set("Location", "/")
+			} else if via == "cli" {
+				w.Header().Set("Location", fmt.Sprintf("/#!/cliauth:s:%s", session.UUID.String()))
```
thomasmmitchell: Can you redirect to an address with a query param?

core/github_auth.go
jhunt: Oops.  Missed that one.

core/api.go
jhunt: Thought about it, but it doesn't hurt anything, and it makes it easier to troubleshoot ex post facto.

core/api.go
jhunt: Yes, but we're redirecting to a URI with a fragment identifier.

jhunt:
@thomasmmitchell feedback addressed, can you take another look?

PR #348, opened 2017-10-13
Provide Diagnostic Messages via API, in debug mode

jhunt:
The new v8 route framework will now issue additional diagnostic message
to API callers when something fails, server-side, if the SHIELD Core
is running in `debug` mode.

Instead of just this:

    {
      "error": "Something went horribly wrong"
    }

API consumers will get this (again, only in `debug` mode):

    {
      "error": "Something went horribly wrong",
      "diagnostic": "server-side error: sql: expected 8 arguments, got 6"
    }

Initially, we did not propagate the exact errors to the front-ends to
avoid issues relating to information disclosure.  However, since we do
not advice (nor support) running production SHIELD Cores in debug mode,
this is not a concern in this case.

What this does net us is additional speed in UI / CLI development, since
we can view the errors that are normally only logged server-side, and
more quickly approach a fix solution.

### DIFF
--- a/core/v2.go
+++ b/core/v2.go
@@ -107,7 +107,9 @@ type v2LocalUser struct {
 }
 
 func (core *Core) v2API() *route.Router {
-	r := &route.Router{}
+	r := &route.Router{
+		Debug: core.debug,
+	}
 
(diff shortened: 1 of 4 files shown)


thomasmmitchell:
llctm

PR #353, opened 2017-10-16
Inherit global retention policies when creating a tenant

xanderrman:
when a tenant is created via cli or web ui/api, the newly created tenant will inherit all of the global templated retention policies created by the site admin.

### DIFF
--- a/db/retention.go
+++ b/db/retention.go
@@ -178,3 +178,18 @@ func (db *DB) DeleteRetentionPolicy(id uuid.UUID) (bool, error) {
 		id.String(),
 	)
 }
+
+//InheritRetentionTemplates gives a tenant the global (templated) retention policies
+func (db *DB) InheritRetentionTemplates(tenantUUID uuid.UUID) error {
+
+	policies, err := db.GetAllRetentionPolicies(&RetentionFilter{ForTenant: "00000000-0000-0000-0000-000000000000"})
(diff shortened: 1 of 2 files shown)


db/retention.go
```diff
@@ -178,3 +178,18 @@ func (db *DB) DeleteRetentionPolicy(id uuid.UUID) (bool, error) {
 		id.String(),
 	)
 }
+
+//InheritRetentionTemplates gives a tenant the global (templated) retention policies
+func (db *DB) InheritRetentionTemplates(tenantUUID uuid.UUID) error {
+
+	policies, err := db.GetAllRetentionPolicies(&RetentionFilter{ForTenant: "00000000-0000-0000-0000-000000000000"})
```
jhunt: Use `uuid.NIL.String()` instead

db/retention.go
```diff
@@ -178,3 +178,18 @@ func (db *DB) DeleteRetentionPolicy(id uuid.UUID) (bool, error) {
 		id.String(),
 	)
 }
+
+//InheritRetentionTemplates gives a tenant the global (templated) retention policies
+func (db *DB) InheritRetentionTemplates(tenantUUID uuid.UUID) error {
```
jhunt: Can we pass the whole Tenant object and just pull the UUID out of it?

PR #354, opened 2017-10-17
Token auth through CLI

thomasmmitchell:
* Token auth provider now does auth on its init/cli endpoint. Its redir
endpoint returns an error and shouldn't be used.
* API library had its auth backend detection updated, and a flow was
added for performing token auth on an auth backend.
* The CLI can now auth to a token provider with a --provider flag (and
an optional --token flag).
* Fixed a minor bug with SHIELD whoami where sysrole wasn't populating.
* fixed testdev config to properly configure token backend
* Added error check for token backend if no tokens were configured

### DIFF
--- a/api/access.go
+++ b/api/access.go
@@ -3,8 +3,9 @@ package api
 import (
 	"bytes"
 	"encoding/json"
-	"io/ioutil"
+	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/pborman/uuid"
 )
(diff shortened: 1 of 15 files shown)


api/access.go
```diff
@@ -102,30 +103,20 @@ func Login(username, password string) (sessionID string, user *AuthIDOutput, err
 		panic("Could not marshal auth struct that we JUST made")
 	}
 
-	req, err := http.NewRequest("POST", uri.String(), bytes.NewReader(j))
+	var req *http.Request
+	req, err = http.NewRequest("POST", uri.String(), bytes.NewReader(j))
```
jhunt: This seems superfluous to me

api/access.go
```diff
@@ -102,30 +103,20 @@ func Login(username, password string) (sessionID string, user *AuthIDOutput, err
...
 	if err != nil {
 		return
 	}
 
-	r, err := makeRequest(req)
+	user = &AuthIDOutput{}
+	var header http.Header
+	header, err = uri.HeaderRequest(user, req)
```
jhunt: Rename `HeaderRequest` to `RequestWithHeaders`

cmd/shield/commands/info/usage.go
```diff
@@ -71,6 +71,7 @@ func printEnvVars() {
 	header("ENVIRONMENT VARIABLES")
 	contents("SHIELD_TRACE\t\tset to 'true' for trace output.")
 	contents("SHIELD_DEBUG\t\tset to 'true' for debug output.")
+	contents("SHIELD_API_TOKEN\t\tset to send API token on requests")
```
jhunt: May want to mention that it should be set to the Token Value itself.  It kind of sounds like its a boolean (i.e. "if set, send the API token!")

core/token_auth.go
```diff
@@ -7,7 +7,9 @@ import (
 
 	"github.com/pborman/uuid"
 
+	"github.com/starkandwayne/goutils/log"
```
jhunt: Minor formatting nit; since goutils is outside of SHIELD, move this import up to be next to pborman up there, not down with the "internal" SHIELD deps down here.

api/access.go
thomasmmitchell: Done if the PR ever takes the commit

api/access.go
thomasmmitchell: Trying not to shadow err

core/token_auth.go
thomasmmitchell: Even better, use the correct log library.

thomasmmitchell: nvm, this is server side. Yeah, I can do that. But I think this might be a goimports thing. I'll see.

PR #357, opened 2017-10-18
Support for storage thresholds

daviddob:
Added storage threshold functionality to the DB, API, and webUI
Altered archive_count to just be an int everywhere (int32 archives should be more than enough)
Pulled out Forecasting/Predicted implementation for now on webUI

### DIFF
--- a/core/v2.go
+++ b/core/v2.go
@@ -1474,10 +1474,11 @@ func (core *Core) v2API() *route.Router {
 	// }}}""
 	r.Dispatch("POST /v2/tenants/:uuid/stores", func(r *route.Request) { // {{{
 		var in struct {
-			Name    string `json:"name"`
-			Summary string `json:"summary"`
-			Agent   string `json:"agent"`
-			Plugin  string `json:"plugin"`
+			Name      string `json:"name"`
+			Summary   string `json:"summary"`
+			Agent     string `json:"agent"`
+			Plugin    string `json:"plugin"`
+			Threshold int    `json:"threshold"`
 
 			Config map[string]interface{} `json:"config"`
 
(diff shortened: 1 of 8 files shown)


web2/htdocs/index.html
```diff
@@ -1097,9 +1097,8 @@
                   </style>
                 </defs>
 
-                [[
-                    var total = Math.max(_.store.used, _.store.threshold, _.store.projected, _.store.total);
-                ]]
+                [[# Base value of 1byte, add 20% padding for spacing ]]
+                [[ var total = Math.max(_.store.used, _.store.threshold, 1) * 1.2; ]]
```
jhunt: Probably ought to bump the minimum to 1MB, since 1b will present as "definitely something wrong with SHIELD" not just "nothing using this and no threshold"

web2/htdocs/js/lib.js
```diff
@@ -987,12 +987,27 @@
 
    ***************************************************/
   exported.bytes = function (x) { // {{{
-    var units = ["b", "K", "M", "G", "T"];
+    var units = ["b", "Kb", "Mb", "Gb", "Tb"];
```
jhunt: Thoughts on capitalizing `b` in the >1k scenarios?  GB, MB, etc.?

core/v2.go
```diff
@@ -1474,10 +1474,11 @@ func (core *Core) v2API() *route.Router {
...
 		var in struct {
-			Name    string `json:"name"`
-			Summary string `json:"summary"`
-			Agent   string `json:"agent"`
-			Plugin  string `json:"plugin"`
+			Name      string `json:"name"`
+			Summary   string `json:"summary"`
+			Agent     string `json:"agent"`
+			Plugin    string `json:"plugin"`
+			Threshold int    `json:"threshold"`
```
jhunt: Threshold should be an int64.

PR #358, opened 2017-10-19
New token auth

thomasmmitchell:


thomasmmitchell:
I'm actually thinking tests will probably fail because of SHIELD_API_TOKEN envvar... hold on.

PR #360, opened 2017-10-23
Updated timespec tanguage to support cardinality

daviddob:
Added "every n hours from (timeOfDay)" where (0 < n < 23) && n is an integer
Added "every half hour from (timeOfDay)
Added "every quarter hour from (timeOfDay)


timespec/timespec_test.go
```diff
@@ -24,6 +24,36 @@ var _ = Describe("Timespec", func() {
...
+		})
+
+		It("can stringify half hour specs", func() {
+			spec := &Spec{
+				Interval:    Hourly,
+				TimeOfDay:   inMinutes(0, 15),
+				Cardinality: 0.5,
+			}
+
+			Ω(spec.String()).Should(Equal("every 0.50 hours from 0:15"))
```
jhunt: This should return "every half hour from 0:15", so that we don't generate a string we cannot parse.

timespec/timespec_test.go
```diff
@@ -24,6 +24,36 @@ var _ = Describe("Timespec", func() {
...
+		})
+
+		It("can stringify quarter hour specs", func() {
+			spec := &Spec{
+				Interval:    Hourly,
+				TimeOfDay:   inMinutes(0, 15),
+				Cardinality: 0.25,
+			}
+
+			Ω(spec.String()).Should(Equal("every 0.25 hours from 0:15"))
```
jhunt: Likewise, this should return "every quarter hour from 0:15", so that we don't generate a string we cannot parse.

PR #363, opened 2017-10-27
Non SiteAdmin Tenant Management

daviddob:
Added UI for management of invitation/banishment for tenant admins who are not necessarily site admins

### DIFF
--- a/core/auth.go
+++ b/core/auth.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/pborman/uuid"
 	"github.com/jhunt/go-log"
+	"github.com/pborman/uuid"
 
 	"github.com/starkandwayne/shield/db"
 	"github.com/starkandwayne/shield/route"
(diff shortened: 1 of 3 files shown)


core/auth.go
```diff
@@ -198,6 +198,38 @@ func (core *Core) hasSystemRole(r *route.Request, roles ...string) bool {
...
 }
 
+func (core *Core) CanManageTenants(r *route.Request, tenant string) bool {
+	user, err := core.AuthenticatedUser(r)
+	if user == nil || err != nil {
+		r.Fail(route.Unauthorized(err, "Authorization required"))
+		return false
+	}
+
+	if user.SysRole == "admin" {
```
jhunt: System Managers are allowed to manage tenants too

daviddob:
Before this gets merged, currently on v8 (and in this current rework) system engineers have the tenants menu in the Admin pane however v2/tenants is a protected endpoint of manager and greater so they get an access denied. Should they have this menu? If so, should v2/tenants be a protected endpoint?

PR #364, opened 2017-10-30
Storage health check

xanderrman:
added proper cloud storage health checking.  SHIELD core will periodically have  test all of the cloud stores, and ensure they can be read from, written to, and that the proper bytes are being written.  The web UI will display a red icon and communicate to the user that cloud storage is failing in the HUD if any of the stores fail the store test, and the db maintains a "healthy" status of each individual store, as well as the UUID of the last store test task that was run, so that an operator may look up further details of the test if a store check returns not healthy


xanderrman:
hehe wrong branch

PR #370, opened 2017-11-03
added auth checks on health endpoints

xanderrman:
if you are not in a tenant and do not have a system role you will not see the hud, and if you attempt to ping the health endpoints without auth you will be rejected


xanderrman:
so when shield runs the routine health ping, if someones not on a tenant/sysrole you want me to dump them to login (maybe with a message)?

PR #372, opened 2017-11-07
Ruggedize SHIELD Web UI

jhunt:
- Interpose nginx in testdev configuration
- Add various security-conscious headers via nginx
- Combine all inline javascript into origin-hosted files
  (for Content-Source-Policy: headers)


dennisjbell:
Deployed and reviewed, only issue found was init.js was being cached, which was resolved with https://github.com/starkandwayne/shield/pull/372/commits/99e4ebc40bd46f53610ae85d0dae2688964666cf

PR #379, opened 2017-12-15
Orange contribution for a Cassandra target plugin

bgandon:
Hi folks,

Here is an Orange FR contribution for a Cassandra target plugin. This was successfully tested with our Work-In-Progress [cassandra release](https://github.com/orange-cloudfoundry/cassandra-cf-service-boshrelease).

Don't hesitate to make any feedback or comments and we'll improve the code and documentation in any way you may find relevant!

Cheers,
Benjamin GANDON

### DIFF
--- a/plugin/cassandra/plugin.go
+++ b/plugin/cassandra/plugin.go
@@ -77,16 +77,16 @@ import (
 )
 
 const (
-	DefaultHost = "127.0.0.1"
-	DefaultPort = "9042"
-	DefaultUser = "cassandra"
+	DefaultHost     = "127.0.0.1"
+	DefaultPort     = "9042"
+	DefaultUser     = "cassandra"
 	DefaultPassword = "cassandra"
-	DefaultBinDir = "/var/vcap/jobs/cassandra/bin"
-	DefaultDataDir = "/var/vcap/store/cassandra/data"
-	DefaultBsdTar = "/var/vcap/packages/bsdtar/bin/bsdtar"
+	DefaultBinDir   = "/var/vcap/jobs/cassandra/bin"
+	DefaultDataDir  = "/var/vcap/store/cassandra/data"
+	DefaultBsdTar   = "/var/vcap/packages/bsdtar/bin/bsdtar"
 
 	VcapOwnership = "vcap:vcap"
-	SnapshotName = "shield-backup"
+	SnapshotName  = "shield-backup"
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


bgandon:
Here I made my tests with the GNU `tar` utility, as provided by Linux Stemcells. I have a question about preferring `bsdtar` over GNU `tar`.

What are the limitations of GNU `tar` that `bsdtar` addresses?

jhunt:
The primary reason for using bsdtar was for the `fs` plugin, which needed to be able to both `--include` and `--exclude` file globs.  Since the `fs` plugin has been rewritten in v8, `bsdtar` is no longer required, although I would strongly encourage you to look at using Golang's `archive/tar` if possible.

bgandon:
Ok, we'll do something similar to the `xtrabackup` plugin.
Unfortunately I won't have time to dig into Golang's `archive/tar`.

plugin/cassandra/plugin.go
```diff
@@ -0,0 +1,666 @@
...
+
+// Called to validate endpoints from the command line
+func (p CassandraPlugin) Validate(endpoint plugin.ShieldEndpoint) error {
+	var (
+		s    string
+		err  error
+		fail bool
+	)
+
+	s, err = endpoint.StringValueDefault("cassandra_host", "")
```
jhunt: As a matter of style, we are trying to move attribute names to be less namespaced.  For example, "host" is clearer than "cassandra_host" -- the original fear that people would just randomly change plugins without re-doing the config has proved to be unfounded.

Would you mind just dropping the `cassandra_` prefix on all configuration options?

plugin/cassandra/plugin.go
```diff
@@ -0,0 +1,666 @@
...
+
+	s, err = endpoint.StringValue("cassandra_keyspace")
+	if err != nil {
+		fmt.Printf("@R{\u2717 cassandra_keyspace      %s}\n", err)
+		fail = true
+	} else {
+		fmt.Printf("@G{\u2713 cassandra_keyspace}      @C{%s}\n", s)
+	}
+
+	s, err = endpoint.StringValueDefault("cassandra_bindir", "")
```
jhunt: In SHIELD v8 and beyond, we now set the `$PATH` environment variable more explicitly, to free up the plugin authors from having to code in explicit path support for executing binaries, and to free up SHIELD users from having to micromanage where they installed all the things.  That responsibility now lies with the team(s) deploying SHIELD agent to the places.

We should therefore remove the `cassandra_bindir` property and just rely on `$PATH`.

See https://github.com/starkandwayne/shield-boshrelease/blob/master/jobs/shield-agent/spec#L49-L52

plugin/cassandra/plugin.go
```diff
@@ -0,0 +1,666 @@
...
+	if err != nil {
+		fmt.Printf("@R{\u2717 cassandra_datadir       %s}\n", err)
+		fail = true
+	} else if s == "" {
+		fmt.Printf("@G{\u2713 cassandra_datadir}       using default @C{%s}\n", DefaultDataDir)
+	} else {
+		fmt.Printf("@G{\u2713 cassandra_datadir}       @C{%s}\n", s)
+	}
+
+	s, err = endpoint.StringValueDefault("cassandra_tar", "")
```
jhunt: Similar to the comments on `cassandra_bindir`, we can just remove this outright and just exec `tar`.  Given that this plugin doesn't need anything special à la bsdtar, standard GNU tar in `/bin` should suffice.

plugin/cassandra/plugin.go
```diff
@@ -0,0 +1,666 @@
...
+	} else {
+		fmt.Printf("@G{\u2713 cassandra_password}      @C{%s}\n", s)
+	}
+
+	s, err = endpoint.StringValue("cassandra_keyspace")
+	if err != nil {
+		fmt.Printf("@R{\u2717 cassandra_keyspace      %s}\n", err)
+		fail = true
+	} else {
+		fmt.Printf("@G{\u2713 cassandra_keyspace}      @C{%s}\n", s)
```
jhunt: Is there any particular format to the keyspace argument?  The documentation comments above imply that all keyspace names should start with `ks` -- if so, we should validate that here.

plugin/cassandra/plugin.go
```diff
@@ -0,0 +1,666 @@
...
+		return err
+	}
+	fmt.Fprintf(os.Stderr, "@G{\u2713 Create new snapshot}\n")
+
+	// Here we need to copy the snapshots/shield-backup directories into a
+	// {keyspace}/{tablename} structure that we'll temporarily put in
+	// /var/vcap/store/shield/cassandra. Then we can tar it all and stream
+	// that to stdout.
+
+	baseDir := "/var/vcap/store/shield/cassandra"
```
jhunt: I would prefer not to use `store` for this; ideally, golang's `archive/tar` library will allow you to stream the archive, with arbitrary paths, without touching the filesystem (which can be very important if you are backing up a thing and don't want to enforce 50% disk usage max).

plugin/cassandra/plugin.go
bgandon: No prob., I'll do that.

plugin/cassandra/plugin.go
bgandon: The `ks` prefix is a convention [enforced by the broker](https://github.com/orange-cloudfoundry/cassandra-cf-service-boshrelease/blob/master/src/cassandra-open-service-broker/src/main/java/com/orange/oss/osb/cassandra/util/Converter.java#L8) in the Cassandra BOSH Release we are currently building. So there's nothing to validate here. The `ksXXX` example is just a suggestion.

plugin/cassandra/plugin.go
bgandon: Makes sense indeed. So we'll provide the `/var/vcap/cassandra/job/bin` directory to the SHIELD agent through this `env.path` list.

plugin/cassandra/plugin.go
bgandon: Yep, no problem.

plugin/cassandra/plugin.go
bgandon: Interesting approach. But the comment is wrong and needs to be updated.

At backup time, we actually hard-link the created immutable snapshot files into a suitable directory structure, that we delete once the tar file is streamed. Cassandra generates the snapshot files only in subdirectories of its own data dir. So, it's already in `/var/vcap/store` and there's nothing we can do about that.

At restore time, the `sstableloader` tool also needs the whole tar archive to be extracted somewhere before it starts working. Here we've chosen to put that on the persistent disk. I'm raising this point here because we could do differently and put that on the ephemeral disk in `/var/vcap/data`. For the time being, we just thought that if we need the extra space at backup time, then we can use it at restore time.

All in all, our current approach would be to document (here in this plugin and in the Cassandra Bosh release) the need for extra space in `/var/vcap/store`. But don't hesitate to suggest any other way of seeing things!

bgandon:
Hi James,
I finished my first answers to your remarks. I'll submit new code where it's needed. No rush on my side though, because I'm currently off and be back on the 2nd week of Jan.
/Benjamin

plugin/cassandra/plugin.go
jhunt: Carry on then 😃

plugin/cassandra/plugin.go
jhunt: Ah, so it's an in-built limitation of the actual backup strategy.  As long as we document that and explain it, I'm fine with it (although I do hate requiring all that extra storage...)

jhunt:
Thanks @bgandon! I will merge once I see new code and/or you nudge me to remember that I said I would merge 😄

bgandon:
Hi James,
First of all, happy new year to you and all Stark & Wayne folks!
I've just submitted changes to the code as we agreed based on your feedback, except the rewrite using Golang's `archive/tar` library.
Is this a blocker for you, or are you OK to accept this PR as is?
/Benjamin

jhunt:
Sounds good to me!  We can circle back and add archive/tar support back in later if anyone wants it bad enough.

PR #381, opened 2017-12-18
Improve backup and restore procedures in xtrabackup plugin

plafosse-orange:
At Orange FR, we backup our MySQL instances via the xtrabackup plugin. It's working properly. But when the backup or the restore fail, we have no trace of the error in the logfile. 

In this PR work, we suggest to make some preliminary tests and to indicate in the log file the result of these tests.
For the backup, we test the existence of the backup temporary directory.

For the restore, we test 
- that the target Mysql instance is stopped
- the existence of the backup temporary directory
- the existence of the MySQL data directory
- and that the MySQL data directory is empty

Also, at the end of restore, we assign the uid-gid couple of the MySQL data directory to the contents of this directory

Don't hesitate to suggest any improvements to this work!

Cheers

### DIFF
--- a/plugin/xtrabackup/plugin.go
+++ b/plugin/xtrabackup/plugin.go
@@ -50,7 +50,7 @@
 // mysql_temp_targetdir:
 // This option specifies the absolute path to a temporary directory used by
 // the `xtrabackup` tool to backup the MySQL databases. It must be empty after
-// each run of the plugin.
+// each run of the plugin.i It must be as big as the estimated MySQL data directory.
 //
 // mysql_tar:
 // This option specifies the absolute path to the `tar` tool.
@@ -64,8 +64,11 @@
 // RESTORE DETAILS
 //
 // To restore, the `xtrabackup` plugin moves back the backed up data files to
-// the MySQL data directory. The MySQL data directory needs to be empty before
-// the restore operation.
+// the MySQL data directory. Before the restore operation, MySQL must be stopped and 
+// the MySQL data directory needs to be empty. 
+//
+// To complete the restore of a Galera cluster, all nodes must be stopped. The previously restored node must 
+// be rebooted in bootstrap mode. The other nodes will be added to the second time to the cluster..
 //
 // DEPENDENCIES
 //
@@ -76,8 +79,9 @@ package main
 
 import (
 	"os"
-
 	fmt "github.com/jhunt/go-ansi"
+	"path/filepath"
+        "syscall"
 
 	"github.com/starkandwayne/shield/plugin"
 )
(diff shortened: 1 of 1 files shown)


plugin/xtrabackup/plugin.go
```diff
@@ -50,7 +50,7 @@
 // mysql_temp_targetdir:
 // This option specifies the absolute path to a temporary directory used by
 // the `xtrabackup` tool to backup the MySQL databases. It must be empty after
-// each run of the plugin.
+// each run of the plugin.i It must be as big as the estimated MySQL data directory.
```
jhunt: You have an extra 'i' here (vim?)

plugin/xtrabackup/plugin.go
```diff
@@ -76,8 +79,9 @@ package main
 
 import (
 	"os"
-
 	fmt "github.com/jhunt/go-ansi"
+	"path/filepath"
+        "syscall"
```
jhunt: You should `go fmt` this file.

Also, we group our imports into three sections:

```
import (
  /* standard library imports */

  /* external libraries */

  /* modules internal to SHIELD (but still on Github) */
)
```

Each group should be separated by blank lines (two `\n`'s)

plugin/xtrabackup/plugin.go
```diff
@@ -302,15 +320,18 @@ func (p XtraBackupPlugin) Backup(endpoint plugin.ShieldEndpoint) error {
 
 	plugin.DEBUG("Executing: `%s`", cmdString)
 	if err = plugin.ExecWithOptions(opts); err != nil {
+		fmt.Fprintf(os.Stderr,"@R{\u2717 Create backup files}\n")
 		return err
 	}
+	fmt.Fprintf(os.Stderr,"@G{\u2713 Create backup files}\n")
```
jhunt: These log message read better as "Creating backup files failed: <err>" and "Created backup files"

plugin/xtrabackup/plugin.go
```diff
@@ -302,15 +320,18 @@ func (p XtraBackupPlugin) Backup(endpoint plugin.ShieldEndpoint) error {
...
 
 	// create and return archive
 	cmdString = fmt.Sprintf("%s -cf - -C %s .", xtrabackup.Tar, targetDir)
 	if err = plugin.Exec(cmdString, plugin.STDOUT); err != nil {
+		fmt.Fprintf(os.Stderr,"@R{\u2717 Create archive}\n")
 		return err
 	}
-
+	fmt.Fprintf(os.Stderr,"@G{\u2713 Create archive}\n")
```
jhunt: These log messages read better as "Creating archive failed: <err>" and "Created archive"

plafosse-orange:
The Xtrabackup plugin is modified with your comments.

jhunt:
LGTM.  Thanks!

PR #382, opened 2017-12-20
Resolved duplicate file paths and inability to write to restored files

daviddob:


plugin/fs/plugin.go
```diff
@@ -192,7 +193,7 @@ func (p FSPlugin) Backup(endpoint plugin.ShieldEndpoint) error {
 			return err
 		}
 
-		header.Name = path
+		header.Name = strings.Replace(path, cfg.BasePath, "", 1)
```
jhunt: Does this handle BasePath's that don't end in `/`, or does it make absolute files in the resulting archive?

jhunt: Pushed a commit to handle the leading slash problem, which takes care of this feedback.  Once travis finishes up tests, I'll merge this.

I found an interesting error when trying to test this with standard bsdtar on my darwin box:

```
tar: Archive entry has empty or unreadable filename ... skipping.
```

It doesn't get any more verbose, so that requires some more direct binary introspection on the generated archive.  That's also an issue for a different PR.

jhunt: Spoke too soon.  That was a regression introduced by this PR.  Now fixed as of 9371c16

PR #385, opened 2018-01-02
Shield Fixed Key Encryption

daviddob:
- Introduces a new boolean for jobs, allowing operators to mark them as "DR" jobs that use a fixed key.
- Introduces a Disaster Key used to derive fixed encryption keys used for fixed-key jobs provided at initialization time.
- On shield rekey, operators will be given the option to rotate their Disaster Key as well; in which the same process from shield init is followed.


client/v2/shield/init.go
```diff
@@ -1,10 +1,32 @@
...
 		Master string `json:"master"`
 	}{
 		Master: master,
 	}
-	return c.post("/v2/init", in, nil)
+
+	out := struct {
+		DisasterKey string `json:"disaster_key"`
+	}{
+		DisasterKey: "",
```
jhunt: You don't need to initialize this anon structure var.

client/v2/shield/init.go
```diff
@@ -1,10 +1,32 @@
...
+
+func (c *Client) SplitKey(s string, n int) string {
+	var buffer bytes.Buffer
+	for i, rune := range s {
+		buffer.WriteRune(rune)
+		if i%n == (n-1) && i != (len(s)-1) {
+			buffer.WriteRune('\n')
+		}
+	}
+	return buffer.String()
```
jhunt: There's a more legible way of doing this (breaking a long string into lines) using string slices; we should investigate.

client/v2/shield/rekey.go
```diff
@@ -1,12 +1,23 @@
...
 		OldMaster: oldmaster,
 		NewMaster: newmaster,
+		RekeyDR:   rekeyDR,
 	}
-	return c.post("/v2/rekey", in, nil)
+
+	out := struct {
+		DisasterKey string `json:"disaster_key"`
+	}{
+		DisasterKey: "",
```
jhunt: You don't need to initialize this anon structure var.

cmd/buckler/main.go
```diff
@@ -83,6 +83,7 @@ var opts struct {
 	Rekey struct {
 		OldMaster string `cli:"--old-master"`
 		NewMaster string `cli:"--new-master"`
+		RekeyDR   bool   `cli:"--rekey-dr"`
```
jhunt: Can we rename this switch to `--rotate-disaster-key`?

cmd/buckler/main.go
```diff
@@ -842,9 +855,19 @@ func main() {
...
+		disasterKey, err := c.Rekey(opts.Rekey.OldMaster, opts.Rekey.NewMaster, opts.Rekey.RekeyDR)
 		bail(err)
 
+		if disasterKey != "" {
+			fmt.Printf("@R{BELOW IS YOUR DISASTER KEY FOR RECOVERING FIXED-KEY BACKUPS.}\n")
+			fmt.Printf("@R{SAVE THIS IN A SECURE LOCATION.}\n")
+			fmt.Printf("----------------------------------------------------------------\n")
+			fmt.Printf("@Y{" + c.SplitKey(disasterKey, 64) + "}")
+			fmt.Printf("\n----------------------------------------------------------------\n")
+		} else if opts.Rekey.RekeyDR {
```
jhunt: Is it possible for Rekey() to return a nil err and an empty disaster key?  If it is, we should fix that.

cmd/buckler/main.go
```diff
@@ -812,11 +815,21 @@ func main() {
...
 
 		fmt.Printf("SHIELD core unlocked successfully.\n")
 
+		if disasterKey != "" {
+			fmt.Printf("@R{BELOW IS YOUR DISASTER KEY FOR RECOVERING FIXED-KEY BACKUPS.}\n")
+			fmt.Printf("@R{SAVE THIS IN A SECURE LOCATION.}\n")
+			fmt.Printf("----------------------------------------------------------------\n")
+			fmt.Printf("@Y{" + c.SplitKey(disasterKey, 64) + "}")
+			fmt.Printf("\n----------------------------------------------------------------\n")
+		} else {
```
jhunt: Is it possible for Initialize() to return a nil err and an empty disaster key? If it is, we should fix that.

cmd/buckler/main.go
```diff
@@ -2103,9 +2126,9 @@ func main() {
 			)
 		*/
 		/* FIXME: support --long / -l and maybe --output / -o "fmt-str" */
-		tbl := tui.NewTable("UUID", "Name", "Summary", "Schedule", "Status", "Policy", "SHIELD Agent", "Target", "Store")
+		tbl := tui.NewTable("UUID", "Name", "Summary", "Schedule", "Status", "Policy", "SHIELD Agent", "Target", "Store", "Disaster Recovery")
 		for _, job := range jobs {
-			tbl.Row(job, job.UUID, job.Name, job.Summary, job.Schedule, job.Status(), job.Policy.Name, job.Agent, job.Target.Name, job.Store.Name)
+			tbl.Row(job, job.UUID, job.Name, job.Summary, job.Schedule, job.Status(), job.Policy.Name, job.Agent, job.Target.Name, job.Store.Name, job.DisasterRecovery)
```
jhunt: What do you think about putting a "Type" column before "Name", and setting it to "normal" if not a DR job, and "DR" if it is (probably with some coloring for boldness), vs. having a dedicated "Disaster Recovery" column?

cmd/buckler/main.go
```diff
@@ -2147,6 +2170,7 @@ func main() {
 		r.Add("Storage Plugin", job.Store.Plugin)
 		r.Break()
 
+		r.Add("Disaster Recovery", strconv.FormatBool(job.DisasterRecovery))
```
jhunt: If we do the "Type" thing in the table, we should do the same here.

db/schema_v4.go
```diff
@@ -87,17 +87,18 @@ func (s v4Schema) Deploy(db *DB) error {
...
+	               store_uuid         UUID NOT NULL,
+	               tenant_uuid        UUID,
+	               schedule           TEXT NOT NULL,
+	               next_run           INTEGER DEFAULT 0,
+	               retention_uuid     UUID NOT NULL,
+	               priority           INTEGER DEFAULT 50,
+	               paused             BOOLEAN,
+	               name               TEXT,
+				   summary            TEXT,
+				   disaster_recovery  INTEGER DEFAULT 0
```
jhunt: You've got some whitespace issues here...

web2/htdocs/index.html
```diff
@@ -243,6 +243,36 @@ <h2>CLI Authentication Credentials</h2>
...
           </div>
     <!-- }}} --></script>
+    <script type="text/html" id="tpl--disasterkey"><!-- {{{ -->
+        [[#
+          {disasterkey} template
+  
+          Displays the Shield Disaster Key, so that operators can restore
+          fixed-key backups in the event of a disaster.
+        ]]
+            <div id="cliauth" class="frontdoor">
```
jhunt: This shouldn't be "cliauth" - probably need to rework CSS if it doesn't fall under `.frontdoor`

web2/htdocs/index.html
```diff
@@ -243,6 +243,36 @@ <h2>CLI Authentication Credentials</h2>
...
+              <div>
+                <h1>SHIELD<span>Production<span></h1>
```
jhunt: Shouldn't `Production` come from our SHIELD env?

web2/htdocs/index.html
```diff
@@ -243,6 +243,36 @@ <h2>CLI Authentication Credentials</h2>
...
+                  <p>Here is your SHIELD Disaster Key:</p>
+  
+                  <pre id="disasterkey">[[= _.disaster_key ]]</pre>
+  
+                  <p id="dr-explain">
+                    You will need to save this key in a secure location
+                    for use in recovery of fixed-key backups in a disaster scenario.
+                    Fixed-key backups must be used to backup shield itself. Once you
+                    leave this screen, there is no way to access this key or corresponding
+                    fixed-key backups unless you have saved it!
```
jhunt: We should either change all the other "Disaster Recovery" references to "Fixed Key", or change these "Fixed Key" references to "Disaster Recovery", for consistency.

web2/htdocs/index.html
```diff
@@ -243,6 +243,36 @@ <h2>CLI Authentication Credentials</h2>
...
+                  </p>
+                  <a href="/"><button id="accept">I Understand</button></a>
```
jhunt: What does the accept button do?

web2/htdocs/index.html
```diff
@@ -1029,6 +1059,15 @@ <h3>Configure a New Protected System</h3>[[# {{{ ]]
             <div id="choose-plugin"></div>
             <div id="configure-plugin"></div>
 
+            <div class="ctl" data-field="fixed-key">
+                <label for="target-fixed-key">Fixed-Key Encryption?</label>
```
jhunt: Make the terminology consistent.

web2/htdocs/shield.css
```diff
@@ -865,7 +865,7 @@ form .ctl label {
   line-height: 20px;
   display: block;
 
-  width: 40%;
+  width: 50%;
```
jhunt: How well does this play with the error reporting on form validation?

cmd/buckler/main.go
daviddob: It should not be possible (at least from the API standpoint always a bad, fail, or they key will exist) these checks were more to ensure future api changes dont introduce a bug that isnt caught quickly.

cmd/buckler/main.go
daviddob: (same as above)

cmd/buckler/main.go
daviddob: Type would make a bit more sense if we had more than two and if we expected operators to use fixed-key backups for more than just shield or a rare other case.

web2/htdocs/index.html
daviddob: Just redirects to the landing page which happens to be same as /systems

web2/htdocs/shield.css
daviddob: Seems to work okay, errors still appear as they used to

PR #395, opened 2018-02-22
Add task pagination to API

Proplex:
This PR adds pagination support for task API requests, as well as hard-limiting the amount of tasks per request. The webui and cli have both been given functionality to support this change, as is follows:

# API Changes
From now on, requests to `/v2/tenants/<UUID>/systems/<UUID>` (and other task-related API calls) will return a soft-limit of 30 tasks. Due to schema limitations, the API may return more than 30 tasks if the last task of the request shares a non-unique `requested_at` timestamp.

To offset API requests, the `?before=<int>` query string can be appended to the API request to get the next 30 tasks that occurred exclusively before `<int>` in chronologically descending order.

# Web UI Changes
A button has been added to the bottom of task pages to allow loading older tasks. 

# CLI Changes
CLI now has a new `--before` modifier that allows for offsetting tasks.  It expects a pretty-date format that's used in `shield tasks` output of `Created at`. The `--limit` modifier will automatically assemble rate-limited (100ms) paginated requests if set to >30.

Example usage includes:
```
shield tasks --before "2018-02-21 16:05:03-0500"
shield tasks --limit 12000
shield tasks --before "2018-02-21 16:05:03-0500"  --limit 12000
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,3 @@ demo/shield-agent/in
 demo/shield-core/in
 demo/shared
 demo/cached
-
-# ignore local text editor project files
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ demo/shield-agent/in
 demo/shield-core/in
 demo/shared
 demo/cached
+
+# ignore local text editor project files
+.vscode
--- a/client/v2/shield/tasks.go
+++ b/client/v2/shield/tasks.go
@@ -13,7 +13,7 @@ type Task struct {
 	Owner       string `json:"owner"`
 	StartedAt   int64  `json:"started_at"`
 	StoppedAt   int64  `json:"stopped_at"`
-	CreatedAt   int64  `json:"requested_at"`
+	RequestedAt int64  `json:"requested_at"`
 	Log         string `json:"log"`
 	OK          bool   `json:"ok"`
 	Notes       string `json:"notes"`
@@ -23,12 +23,12 @@ type Task struct {
 }
 
 type TaskFilter struct {
-	Status            string `qs:"status"`
-	Active            *bool  `qs:"active:t:f"`
-	Debug             *bool  `qs:"debug:t:f"`
-	Limit             *int   `qs:limit`
-	Target            string `qs:"target"`
-	BeforeCreatedDate int64  `qs:"before"`
+	Status string `qs:"status"`
+	Active *bool  `qs:"active:t:f"`
+	Debug  *bool  `qs:"debug:t:f"`
+	Limit  *int   `qs:limit`
+	Target string `qs:"target"`
+	Before int64  `qs:"before"`
 }
 
 func fixupTaskResponse(p *Task) {
--- a/client/v2/shield/tasks.go
+++ b/client/v2/shield/tasks.go
@@ -13,6 +13,7 @@ type Task struct {
 	Owner       string `json:"owner"`
 	StartedAt   int64  `json:"started_at"`
 	StoppedAt   int64  `json:"stopped_at"`
+	CreatedAt   int64  `json:"requested_at"`
 	Log         string `json:"log"`
 	OK          bool   `json:"ok"`
 	Notes       string `json:"notes"`
@@ -22,11 +23,12 @@ type Task struct {
 }
 
 type TaskFilter struct {
-	Status string `qs:"status"`
-	Active *bool  `qs:"active:t:f"`
-	Debug  *bool  `qs:"debug:t:f"`
-	Limit  *int   `qs:limit`
-	Target string `qs:"target"`
+	Status            string `qs:"status"`
+	Active            *bool  `qs:"active:t:f"`
+	Debug             *bool  `qs:"debug:t:f"`
+	Limit             *int   `qs:limit`
+	Target            string `qs:"target"`
+	BeforeCreatedDate int64  `qs:"before"`
 }
 
 func fixupTaskResponse(p *Task) {
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -279,13 +279,13 @@ var opts struct {
 	/* }}} */
 	/* TASKS {{{ */
 	Tasks struct {
-		Status            string `cli:"-s, --status"`
-		Active            bool   `cli:"--active"`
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -58,3 +58,6 @@ demo/shield-agent/in
 demo/shield-core/in
 demo/shared
 demo/cached
+
+# ignore local text editor project files
+.vscode
```
jhunt: this should go in your personal .gitignore, not the projects.

client/v2/shield/tasks.go
```diff
@@ -22,11 +23,12 @@ type Task struct {
...
-	Active *bool  `qs:"active:t:f"`
-	Debug  *bool  `qs:"debug:t:f"`
-	Limit  *int   `qs:limit`
-	Target string `qs:"target"`
+	Status            string `qs:"status"`
+	Active            *bool  `qs:"active:t:f"`
+	Debug             *bool  `qs:"debug:t:f"`
+	Limit             *int   `qs:limit`
+	Target            string `qs:"target"`
+	BeforeCreatedDate int64  `qs:"before"`
```
jhunt: Can we rename this to field to just `Before`?

cmd/shield/main.go
```diff
@@ -278,12 +279,13 @@ var opts struct {
...
-		All      bool   `cli:"-a, --all"`
-		Target   string `cli:"--target"`
-		Limit    int    `cli:"-l, --limit"`
+		Status            string `cli:"-s, --status"`
+		Active            bool   `cli:"--active"`
+		Inactive          bool   `cli:"--inactive"`
+		All               bool   `cli:"-a, --all"`
+		Target            string `cli:"--target"`
+		Limit             int    `cli:"-l, --limit"`
+		BeforeCreatedDate string `cli:"--before"`
```
jhunt: Rename to just `Before`, since it mirrors the flag name.

cmd/shield/main.go
```diff
@@ -2635,25 +2633,55 @@ func main() {
...
+
+				tasks = append(tasks, sometasks...)
+				timeBefore = tasks[len(tasks)-1].CreatedAt
+				time.Sleep(100 * time.Millisecond)
+			}
 		}
 
 		/* FIXME: support --long / -l and maybe --output / -o "fmt-str" */
-		tbl := tui.NewTable("UUID", "Type", "Status", "Owner", "Started at", "Completed at")
+		tbl := tui.NewTable("UUID", "Type", "Status", "Owner", "Created at", "Started at", "Completed at")
```
jhunt: I believe in `--json` format, this is `requested_at`, so we should make that match this column header, or make the column header match the JSON output.  Investigate and fix if necessary.

core/v2.go
```diff
@@ -2293,22 +2332,38 @@ func (core *Core) v2API() *route.Router {
...
 			return
 		}
 
-		tasks, err := core.DB.GetAllTasks(
-			&db.TaskFilter{
+		// check to see if we're offseting task requests
+		paginationDate, err := strconv.ParseInt(r.Param("before", "0"), 10, 64)
+
+		var filter db.TaskFilter
+		if paginationDate > 0 {
```
jhunt: Do you actually need the if?  if paginationDate is 0, there's no difference in the filter assignment; `BeforeCreatedDate: 0` is the same as not setting the field, since the zero value of an integer is 0.  We can probably remove the conditional split, and save the duplicated code.

db/tasks.go
```diff
@@ -59,16 +59,18 @@ type TaskInfo struct {
...
+	SkipInactive      bool
+	OnlyRelevant      bool
+	ForOp             string
+	ForTenant         string
+	ForTarget         string
+	ForStatus         string
+	ForArchive        string
+	Limit             int
+	CreatedDate       int64
+	BeforeCreatedDate int64
```
jhunt: Rename to just `Before`

db/tasks.go
```diff
@@ -59,16 +59,18 @@ type TaskInfo struct {
...
+	SkipActive        bool
+	SkipInactive      bool
+	OnlyRelevant      bool
+	ForOp             string
+	ForTenant         string
+	ForTarget         string
+	ForStatus         string
+	ForArchive        string
+	Limit             int
+	CreatedDate       int64
```
jhunt: Instead of using the `Date` suffix, we normally use `At`, to signify that its a timestamp.

db/tasks.go
```diff
@@ -229,12 +241,12 @@ func (db *DB) CreateBackupTask(owner string, job *Job) (*Task, error) {
 		`INSERT INTO tasks
 		    (uuid, owner, op, job_uuid, status, log, requested_at,
 		     store_uuid, store_plugin, store_endpoint,
-			 target_uuid, target_plugin, target_endpoint, restore_key, 
+			 target_uuid, target_plugin, target_endpoint, restore_key,
```
jhunt: Can you remove these spurious whitespace changes from your commit?  We can make a cleanup commit later, but they don't really belong in this changeset.

Proplex: Sorry about that, that's what I get for not using `git diff`.

cmd/shield/main.go
```diff
@@ -2967,7 +3000,7 @@ func main() {
 			break
 		}
 
-		tbl := tui.NewTable("UUID", "Account", "Created At", "Last Seen", "IP Address", "User Agent")
+		tbl := tui.NewTable("UUID", "Account", "Requested At", "Last Seen", "IP Address", "User Agent")
```
jhunt: One too far; this should remain Created At, because this is auth code / sessions stuff.

core/v2.go
```diff
@@ -767,17 +767,56 @@ func (core *Core) v2API() *route.Router {
...
 			archives[archive.UUID.String()] = archive
 		}
+		// check to see if we're offseting task requests
+		paginationDate, err := strconv.ParseInt(r.Param("before", "0"), 10, 64)
 
-		tasks, err := core.DB.GetAllTasks(
-			&db.TaskFilter{
+		var filter db.TaskFilter
+		if paginationDate > 0 {
+			filter = db.TaskFilter{
```
jhunt: Was this one missed?  Should be able to unify the filter object creation and not need the if.

PR #398, opened 2018-03-01
Add the ability to restore from previous SHIELD on initialization

Proplex:
shield

When an operator starts a fresh SHIELD instance (in the case of catastrophic failure, etc.) a new initialization screen is displayed that allows the upload of a previous SHIELD self-backup. An out-of-band restore task is fired, and then shield returns to the login screen.

A few caveats to note: this screen replaces the default initialization page. This means in the case of Vault being blown away, but SHIELD has not been touched, the user is still given the option to restore from a previous SHIELD state (though they can simply just enter a new master password in and be given a fixed key as per previous logic)

This task assumes that SHIELD data is located in `/var/vcap/store/shield/`. It was assumed the use-case of SHIELD would be in a BOSH release.

The ability for the FS plugin to "retrieve" files has been re-added to support this feature. All other "store" functionality stills remains unimplemented as FS remains a plugin unintended to store data.

### DIFF
--- a/core/v2.go
+++ b/core/v2.go
@@ -4,7 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -2589,25 +2592,31 @@ func (core *Core) v2API() *route.Router {
 	// }}}
 
 	/* This following route is an out-of-band task that self-restores SHIELD if chosen
-	to during the initialization of SHIELD */
+	to during the initialization of SHIELD. On success, it stops SHIELD and the web UI
+	waits for a monit restart. On failure, SHIELD will nuke its working directory and crash.
+	Monit will then restart SHIELD, SHIELD will reinitialize, and the operator can try
+	again. */
 	r.Dispatch("POST /v2/bootstrap/restore", func(r *route.Request) { // {{{
 		if core.IsNotSystemAdmin(r) {
 			return
 		}
-		file, header, err := r.Req.FormFile("archive")
+		file, _, err := r.Req.FormFile("archive")
 
-		fmt.Printf(header.Filename)
-
-		f, err := os.Create("/tmp/bootstrap.restore")
+		backupStore, err := ioutil.TempFile("", "SHIELDrestoreBOOTSTRAP")
 		if err != nil {
-			panic(err)
+			r.Fail(route.Oops(err, "Unable to save backup file to disk."))
+			return
 		}
-		defer f.Close()
 
-		io.Copy(f, file)
+		defer backupStore.Close()
+		defer os.Remove(backupStore.Name())                          //clean up tempfile
+		backupPath, backupName := filepath.Split(backupStore.Name()) //necessary for task setup
 
-		if err := f.Sync(); err != nil {
-			panic(err)
+		io.Copy(backupStore, file)
+
+		if err := backupStore.Sync(); err != nil {
+			r.Fail(route.Oops(err, "Unable to save backup file to disk."))
+			return
 		}
 
 		stdout := make(chan string, 1)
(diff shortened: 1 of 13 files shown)


core/v2.go
```diff
@@ -2586,6 +2588,57 @@ func (core *Core) v2API() *route.Router {
...
+	to during the initialization of SHIELD */
+	r.Dispatch("POST /v2/bootstrap/restore", func(r *route.Request) { // {{{
+		if core.IsNotSystemAdmin(r) {
+			return
+		}
+		file, header, err := r.Req.FormFile("archive")
+
+		fmt.Printf(header.Filename)
+
+		f, err := os.Create("/tmp/bootstrap.restore")
```
jhunt: See if we can use a tempfile library like https://golang.org/pkg/io/ioutil/#TempFile

core/v2.go
```diff
@@ -2586,6 +2588,57 @@ func (core *Core) v2API() *route.Router {
...
+		if err != nil {
+			panic(err)
```
jhunt: Don't panic in API calls, ever.

core/v2.go
```diff
@@ -2586,6 +2588,57 @@ func (core *Core) v2API() *route.Router {
...
+				log.Errorf(s)
+			}
+		}()
+
+		/* out-of-band task run to restore SHIELD. /var/vcap/store/shield is hardcoded,
+		as it's assumed that this is a BOSH release */
+		err2 := core.agent.Run("127.0.0.1:5444", stdout, stderr, &AgentCommand{
+			Op:             db.RestoreOperation,
+			TargetPlugin:   "fs",
+			TargetEndpoint: "{\"base_dir\":\"/var/vcap/store/shield/\",\"bsdtar\":\"bsdtar\"}",
```
jhunt: We should pull this from the SHIELD configuration.

core/v2.go
```diff
@@ -2586,6 +2588,57 @@ func (core *Core) v2API() *route.Router {
...
+			StorePlugin:    "fs",
+			StoreEndpoint:  "{\"base_dir\": \"/tmp/\", \"bsdtar\": \"bsdtar\"}",
```
jhunt: This will be based on the Tempfile() from above.

core/v2.go
```diff
@@ -2586,6 +2588,57 @@ func (core *Core) v2API() *route.Router {
...
+			RestoreKey:     "bootstrap.restore",
+			EncryptType:    "aes256-ctr",
+			EncryptKey:     core.vault.ASCIIHexEncode(r.Req.FormValue("fixedkey")[:32], 4),
+			EncryptIV:      core.vault.ASCIIHexEncode(r.Req.FormValue("fixedkey")[32:], 4),
+		})
+
+		if err2 != nil {
+			r.Fail(route.Oops(err2, "SHIELD Restore Failed. You may be in a broken state."))
```
jhunt: Let's talk to Dave about some stuff related to BOSH / re-exec and cleanup operations.

plugin/fs/plugin.go
```diff
@@ -19,7 +19,7 @@ func main() {
 		Version: "1.0.0",
 		Features: plugin.PluginFeatures{
 			Target: "yes",
-			Store:  "no",
+			Store:  "yes",
```
jhunt: We don't want to advertise that we do store ops.  Probably best to leave a comment that we don't advertise, but do implement the Retrieve endpoint, and why.

plugin/fs/plugin.go
```diff
@@ -282,7 +282,23 @@ func (p FSPlugin) Store(endpoint plugin.ShieldEndpoint) (string, int64, error) {
...
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	if _, err = io.Copy(os.Stdout, f); err != nil {
+		return err
+	}
+
+	return nil
```
jhunt: you can just `return io.Copy(os.Stdout, f)`

core/v2.go
Proplex: 🎵it all comes tumbling down, tumbling down, tumbling dooooown🎵

jhunt:
From our discussion:

- Modify the SHIELD core configuration to take a `data_root` option instead of the standalone files.
  - Remove the database file path configuration
  - Remove the vault keys file path configuration
- If the bootstrap restore fails, delete the entire `data_root` and crash the process so monit restarts it
- If the bootstrap restore succeeds, shutdown and let monit restart the process, so that we pick up schema changes if we restored and older version of the database.
- Add a new endpoint that lets us pick up the log of the last failed bootstrap recovery
- Integrate the last-log endpoint into the UI, so that I can see the fail log from the front-end.

Proplex:
Looks like Travis & the dev setup scripts need to be changed to reflect the new config. I'll get on that.

plugin/fs/plugin.go
Proplex: By doing this, aren't we playing hot-potato with the error?

jhunt:
Excellent!

jhunt:
Once Travis agrees, I'll merge this.

PR #400, opened 2018-03-16
Move SHIELD encryption to a standalone binary

Proplex:
After manual testing and debugging, we've discovered that SHIELD was not encrypting the backups it was taking. This was due to two issues: `shield-pipe` was not properly reading the appropriate encryption-related environment variables, but also the change in plugin `Exec()` structure allowed some plugins, such as `fs`, to skip encryption wholesale.

This commit solves both problems, by fixing `shield-pipe` to properly test environment variables, and by removing the encryption/decryption-related code from `Exec()` and into a standalone `shield-crypt` binary that takes the compressed backup from a pipe to `--encrypt` or `--decrypt`.


crypter/shield-crypt/shield-crypt.go
```diff
@@ -0,0 +1,82 @@
+package main
```
jhunt: Can you move this into `cmd/shield-crypt` so that it follows the pattern of all the other binaries?

crypter/shield-crypt/shield-crypt.go
```diff
@@ -0,0 +1,82 @@
...
+	DecryptFull  bool `cli:"--decrypt"`
+}
+
+func main() {
+	var encStream, decStream cipher.Stream
+	var data struct {
+		EncryptionKey  string `json:"enc_key"`
+		EncryptionIV   string `json:"enc_iv"`
+		EncryptionType string `json:"enc_type"`
+	}
```
jhunt: For clarity, rename the structure to `crypt` and drop the `Encryption` prefix on all the members.  This let's us deal with `crypt.Key`, `crypt.IV`, etc.

crypter/shield-crypt/shield-crypt.go
```diff
@@ -0,0 +1,82 @@
...
+
+	decoder := json.NewDecoder(os.NewFile(uintptr(3), "encConfig"))
```
jhunt: Let's name the file handle something like "<3"

crypter/shield-crypt/shield-crypt.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+		} else {
+			if _, err := io.Copy(os.Stdout, os.Stdin); err != nil {
+				panic(err)
+			}
+			os.Exit(0)
+		}
+	}
+
+	var opts CLIOpts
```
jhunt: How do you feel about making this an inline structure definition?

crypter/shield-crypt/shield-crypt.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/jhunt/go-cli"
+	"github.com/starkandwayne/shield/crypter"
+)
+
+type CLIOpts struct {
+	EncryptShort bool `cli:"-e"`
+	DecryptShort bool `cli:"-d"`
+	EncryptFull  bool `cli:"--encrypt"`
+	DecryptFull  bool `cli:"--decrypt"`
```
jhunt: `go-cli` lets you do this:

```
struct {
  Encrypt bool `cli:"-e, --encrypt"`
  Decrypt bool `cli:"-d, --decrypt"`
}
```

We should probably make use of that to simplify handling in main.

crypter/shield-crypt/shield-crypt.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	var opts CLIOpts
+
+	_, _, err := cli.Parse(&opts)
+
+	if err != nil {
+		panic(err)
+	}
+
```
jhunt: What happens if I (pathologically) run `shield-crypt -ed`?

Proplex: I added an expression that checks to see if both flags are set. If so, it'll `os.Exit(1)` with an error message.

PR #405, opened 2018-04-10
Add "Errored" state when `/v2/health` throws an error

Proplex:
Currently, when the backend is unable to get the status of certain services (e.g. Vault), the endpoint throws a 500 and the web UI fails to render (it's expecting a proper health response).

There's a lot of different ways to fix this, including revamping the health check to throw specific "core" modes (e.g. "vault_down") to pass to the API, but that's a rabbit hole and maybe not worth much in the eyes of an operator. Monit should typically bring vault right back up, so we just want a transient error for the minute in case an operator checks during that exact window.

This solution introduces a hard-coded "ERRORED" state on a 500 response. It sets all health values to false and empty data storage arrays. It's designed to inform the operator that SHIELD is not in a good state, but avoids being overly verbose.

shield


jhunt:
Can we change the verbiage to "failing" / "FAILING"?

Proplex:
One commit, hot off the press

shield

PR #406, opened 2018-04-11
add mysql socket to xtrabackup configuration options

JamesClonk:
This will add the possibility to configure the path to the mysql socket in the xtrabackup plugin.
Without it xtrabackup might not find the running mysql process and cannot perform the backup, if the socket is in an unexpected location.

signed-off-by: Fabio Berchtold <fabio.berchtold@swisscom.com>


jhunt:
LGTM!

PR #420, opened 2018-06-28
Adds backup/restore for `safe` target:

dennisjbell:
Backs up the entire secret/... path.

Restores from backup, but keeps any value in secret/vault/seal/keys so
that outdated seal keys do not overwrite new ones.

Takes the following endpoint attributes:
+  vault_url: The url address of your Vault server, including the protocol - example: "https://safe.myorg.mycompany.com"
+ auth_token:  The auth token for a user with privileges to read and write the entire secret/ tree
+ skip_ssl_validation: If using self-signed SSL certificates, set to true.  Defaults to false.

### DIFF
--- a/plugin/safe/plugin.go
+++ b/plugin/safe/plugin.go
@@ -68,12 +68,6 @@ func main() {
 
 type SafePlugin plugin.PluginInfo
 
-type SafeConnectionInfo struct {
-	URL            string
-	Token          string
-	SkipValidation bool
-}
-
 func (p SafePlugin) Meta() plugin.PluginInfo {
 	return plugin.PluginInfo(p)
 }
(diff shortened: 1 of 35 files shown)


plugin/safe/plugin.go
```diff
@@ -0,0 +1,276 @@
...
+	if skipSslValidation {
+		plugin.DEBUG("Skipping SSL validation")
+	}
+
+	return &SafeConnectionInfo{
+		URL:            url,
+		Token:          token,
+		SkipValidation: skipSslValidation,
+	}, nil
+}
```
jhunt: Combine safeConnect with this function, and call it `connect(endpoint)`

daviddob:
Would it be worthwhile to make the root to backup/restore a parameter or should that be added in a future PR? It was a recent addition, but we now have a concourse attached to vault under the `concourse` root instead of `secret`.

jhunt:
@daviddob that's something we can revisit at a later date.

PR #431, opened 2018-07-13
Added software-properties-common in order to install php on aegis

nbari:
`make demo` broken, need to add `ppa:ondrej/php` in order to make it work on `agegis Dockerfile`


jhunt:
Thanks!

PR #432, opened 2018-07-13
Improve info and health endpoints

dennisjbell:
This PR updates the /v2/info and /v2/health endpoints, stabilizing version population and removing FQDN.  It also updates the shield cli and client library to convey the information to the end user more reliably.


dennisjbell:
[Header shows version as (development) in released versions](https://trello.com/c/zgUB575I/66-header-shows-version-as-development-in-released-versions)

PR #434, opened 2018-07-16
vault 0.10.* in dev/testdev

nbari:
add vault version `0.10.*` to allow running `make dev`

### DIFF
--- a/bin/testdev
+++ b/bin/testdev
@@ -84,6 +84,7 @@ preflight() {
   case $v in
   v0.8.*) ;;
   v0.9.*) ;;
+  v0.10.*) ;;
   *) echo
      echo "The installed version of vault ($v) has not been verified"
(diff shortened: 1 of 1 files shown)


dmolik:
Looks good to me, thank you!

jhunt:
Umm... are we 100% sure that Vault 0.10.0's key-value backend will work with SHIELD?  There was a reason we pinned at 0.9.6 and below.

daviddob:
I think this is due to the fact that SHIELD isn't using --dev for the vault and instead is using file backed storage. The new kv backend isn't automatically used as default outside of dev mode yet (not sure when they will make it default for non-dev servers or if they'll just wait till `secret/` isn't a default mount anymore. Breaking change planned for Vault 0.12.X - `secret/` will not be mounted by default at init time).

PR #438, opened 2018-07-31
Begin work on Backblaze B2 plugin

dmolik:
I've started work on supporting Backblaze as an optional storage plugin. I could use some feedback.


jhunt:
I think this looks good; can we rename the plugin to `backblaze` and also write some documentation for it?

dmolik:
Crap, let me clean up the git history.

PR #449, opened 2018-09-05
fix mysql plugin to accept database name when restoring

nbari:
allows using option `mysql_database`

### DIFF
--- a/plugin/mysql/plugin.go
+++ b/plugin/mysql/plugin.go
@@ -272,7 +272,7 @@ func connectionString(info *MySQLConnectionInfo, backup bool) string {
 			db = "--all-databases"
 		}
 	} else {
-		db = ""
+		db = info.Database
 	}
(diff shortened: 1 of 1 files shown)


jhunt:
Can you elaborate on the reason for this change?

nbari:
Mainly to prevent this error `Error No database selected` and therefore allows  to restore the backups

Previous versions of the plugin where like this:

    var db string
    if info.Database != "" {
        db = info.Database
    } else if backup {
        db = "--all-databases"
    }

This helped to pouplate the 

[return fmt.Sprintf("%s -h %s -P %s -u %s", db, info.Host, info.Port, info.User)](https://github.com/starkandwayne/shield/blob/master/plugin/mysql/plugin.go#L277)

With the defined database.

jhunt:
Sounds good to me!

PR #450, opened 2018-09-10
fix minor copy/paste error in "usage" line

red-trela:


jhunt:
I went ahead and fixed this out-of-band, along with two other commands that had similar issues with their USAGE: lines.

PR #451, opened 2018-09-10
Fix linting issues

mastertinner:


mastertinner:
@jhunt, closing without comment?

PR #466, opened 2018-10-18
Update outputs from 'shield target' and shield update-target'

ahartpence:
Currently, when viewing a target in the shield CLI, it does not give any information about the configuration for said target. This PR adds a `Configuration` line to the output of `target` and `update-target` to show the operator the current or the updated configuration without having to run `targets`


jhunt:
:+1:

PR #467, opened 2018-10-19
Make S3 region configurable

### LINKED ISSUE #465: S3 plugin appears to be broken with Swiftstack S3 server using auth v4
When attempting to run an ad hoc backup using verified S3 configuration (auth URI and credentials) and AWS Signature Version 4, the following error is returned:

The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'US' (AuthorizationHeaderMalformed) [raw 
AuthorizationHeaderMalformed
The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'US'txaacbdb1d22ae48bb9f403-005bc8a71aUS]

jhunt:
This makes life easier on some S3 work-alikes, which can require that
S3 operations be given region names that do _not_ correspond to those
used by AWS itself.

Fixes #465

### DIFF
--- a/plugin/s3/plugin.go
+++ b/plugin/s3/plugin.go
@@ -20,6 +20,7 @@ import (
 
 const (
 	DefaultS3Host              = "s3.amazonaws.com"
+	DefaultRegion              = "us-east-1"
 	DefaultPrefix              = ""
 	DefaultSigVersion          = "4"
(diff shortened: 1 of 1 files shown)


krutten:
Changes look good.  Defaults remain the same if not configure.

Issue #463, opened 2018-10-17
The PUT call to create folder and file at same time doesn't work with apache webdav


cwb124:
Getting error when trying to backup to Webdav served by apache.

[Wed Oct 17 19:45:27.164530 2018] [dav:error] [pid 10034] (2)No such file or directory: [client 10.131.160.6:49096] An error occurred while opening a resource.  [409, #0]
[Wed Oct 17 19:47:02.156336 2018] [dav:error] [pid 10176] [client 10.131.160.6:49326] Unable to PUT new contents for /webdav/2018/10/17/2018-10-17-194702-1c512448-ce00-4d9b-bb3b-f536bd41e939.  [403, #0]

Workaround was to create the nested directories first and then backup began successfully.

PR #470, opened 2018-10-24
Add support for MKCOL for Apache WebDAV

krutten:
Nginx WebDAV has the `create_full_put_path` to automatically create required subdirectories.  Apache mod_dav does not have an equivalent option.  Added a MkCol (MakeCollection) method that Put now uses to pre-create any directories needed.


krutten:
Duplicate of #463

PR #474, opened 2018-12-07
Update cassandra pluggin v8

OMI69-orange:
The plugin for cassandra was tested succefully. We need that plugin for our integration environnement. Can you merge it into the master?
regards

### DIFF
--- a/plugin/cassandra/plugin.go
+++ b/plugin/cassandra/plugin.go
@@ -90,106 +90,151 @@ package main
 import (
 	"os"
 	"path/filepath"
+	"sorts"
 	"strings"
 
-	fmt "github.com/jhunt/go-ansi"
+	fmt "github.com/jhunt/go-ansi" // est un alias
 
 	"github.com/starkandwayne/shield/plugin"
 )
 
+// Default configuration values for the plugin
 const (
-	DefaultHost     = "127.0.0.1"
-	DefaultPort     = "9042"
-	DefaultUser     = "cassandra"
-	DefaultPassword = "cassandra"
-	DefaultDataDir  = "/var/vcap/store/cassandra/data"
+	DefaultHost      = "127.0.0.1"
+	DefaultPort      = "9042"
+	DefaultUser      = "cassandra"
+	DefaultPassword  = "cassandra"
+	DefaultSaveUsers = true
+	DefaultBinDir    = "/var/vcap/jobs/cassandra/bin"
+	DefaultDataDir   = "/var/vcap/store/cassandra/data"
 
 	VcapOwnership = "vcap:vcap"
 	SnapshotName  = "shield-backup"
+)
+
+// Array or slices aren't immutable by nature; you can't make them constant
+var (
+	DefaultExcludeKeyspaces = []string{"system_schema", "system_distributed", "system_auth", "system", "system_traces"}
+	SystemAuthTables        = []string{"roles", "role_permissions", "role_members", "resource_role_permissons_index"}
 )
 
 func main() {
 	p := CassandraPlugin{
 		Name:    "Cassandra Backup Plugin",
 		Author:  "Orange",
-		Version: "0.1.0",
+		Version: "1.0.0",
 		Features: plugin.PluginFeatures{
 			Target: "yes",
 			Store:  "no",
 		},
 		Example: `
 {
-  "host"         : "127.0.0.1",      # optional
-  "port"         : "9042",           # optional
-  "user"         : "username",
-  "password"     : "password",
-  "keyspace"     : "db",
-  "datadir"      : "/path/to/data"   # optional
+  "host"              : "127.0.0.1",      # optional
+  "port"              : "9042",           # optional
+  "user"              : "username",
+  "password"          : "password",
+  "include_keyspaces" : ["db"],
+  "exclude_keyspaces" : ["system"],
+  "save_users"        : true,
+  "bindir"            : "/path/to/bin",   # optional
+  "datadir"           : "/path/to/data",  # optional
 }
 `,
 		Defaults: `
 {
-  "host"     : "127.0.0.1",
-  "port"     : "9042",
-  "user"     : "cassandra",
-  "password" : "cassandra",
-  "datadir"  : "/var/vcap/store/cassandra/data"
+  "host"              : "127.0.0.1",
+  "port"              : "9042",
+  "user"              : "cassandra",
+  "password"          : "cassandra",
+  "exclude_keyspaces" : [ "system_schema", "system_distributed", "system_auth", "system", "system_traces" ],
+  "save_users"        : true,
+  "bindir"            : "/var/vcap/jobs/cassandra/bin",
+  "datadir"           : "/var/vcap/store/cassandra/data"
 }
 `,
 		Fields: []plugin.Field{
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "host",
 				Type:     "string",
 				Title:    "Cassandra Host",
 				Help:     "The hostname or IP address of your Cassandra server.",
 				Default:  "127.0.0.1",
-				Required: true,
+				Required: false,
 			},
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "port",
 				Type:     "port",
 				Title:    "Cassandra Port",
 				Help:     "The 'native transport' TCP port that Cassandra server is bound to, listening for incoming connections.",
 				Default:  "9042",
-				Required: true,
+				Required: false,
 			},
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "user",
 				Type:     "string",
 				Title:    "Cassandra Username",
 				Help:     "Username to authenticate to Cassandra as.",
 				Default:  "cassandra",
-				Required: true,
+				Required: false,
 			},
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "password",
 				Type:     "password",
 				Title:    "Cassandra Password",
 				Help:     "Password to authenticate to Cassandra as.",
 				Default:  "cassandra",
-				Required: true,
+				Required: false,
 			},
-			plugin.Field{
+			{
 				Mode:     "target",
-				Name:     "keyspace",
-				Type:     "string",
-				Title:    "Keyspace to Backup",
-				Help:     "The name of the keyspace to backup.",
-				Example:  "system",
-				Required: true,
+				Name:     "include_keyspaces",
+				Type:     "array",
+				Title:    "Keyspaces to Include in the Backup",
+				Help:     "The name of the keyspace to include in the backup.",
+				Example:  "['system']",
+				Required: false,
 			},
-			plugin.Field{
+			{
+				Mode:     "target",
+				Name:     "exclude_keyspaces",
+				Type:     "array",
+				Title:    "Keyspaces to Exclude from Backup",
+				Help:     "The name of the keyspace to exclude from backup.",
+				Example:  "[system']",
+				Default:  "['system_schema', 'system_distributed', 'system_auth', 'system', 'system_traces']",
+				Required: false,
+			},
+			{
+				Mode:     "target",
+				Name:     "save_users",
+				Type:     "boolean",
+				Title:    "to restore users and permissions along with the keyspaces that are restored",
+				Help:     "The name of the keyspace to exclude from backup.",
+				Example:  "true or false",
+				Required: false,
+			},
+			{
+				Mode:     "target",
+				Name:     "bindir",
+				Type:     "abspath",
+				Title:    "Path to binaries nodetool, sstableloader and cqlsh needed for the backup",
+				Help:     "where  to find binaries nodetool, sstableloader and cqlsh",
+				Example:  "true or false",
+				Default:  "/var/vcap/jobs/cassandra/bin",
+				Required: false,
+			},
+			{
 				Mode:     "target",
 				Name:     "datadir",
 				Type:     "abspath",
 				Title:    "Path to Cassandra data/ directory",
 				Help:     "The absolute path to the data/ directory that contains the Cassandra database files.",
 				Default:  "/var/vcap/store/cassandra/data",
-				Required: true,
+				Required: false,
 			},
 		},
 	}
@@ -200,12 +245,16 @@ func main() {
 type CassandraPlugin plugin.PluginInfo
 
 type CassandraInfo struct {
-	Host     string
-	Port     string
-	User     string
-	Password string
-	Keyspace string
-	DataDir  string
+	Host             string
+	Port             string
+	User             string
+	Password         string
+	IncludeKeyspaces []string
+	ExcludeKeyspaces []string
+	SaveUsers        bool
+	//BinDir           string  no need in Shield V8
+	BinDir  string
+	DataDir string
 }
 
 // This function should be used to return the plugin's PluginInfo, however you decide to implement it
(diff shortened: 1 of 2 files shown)


patrick1378:
So I've been doing some testing on this the last few days and one concern I have is that it seems backups created with the old plugin do not seem compatible with the update. We have some thoughts on how we could handle different versions with shield but I wanted to reach out and get your thoughts.

plugin/cassandra/plugin.go
```diff
@@ -485,67 +642,136 @@ func hardLinkAll(src_dir string, dst_dir string) (err error) {
...
+	sort.Strings(cassandra.ExcludeKeyspaces)
+	savedKeyspaces := computeSavedKeyspaces(cassandra.IncludeKeyspaces, cassandra.ExcludeKeyspaces)
+
+	// TODO: here we should extract only the necessary keyspaces
+	cmd = fmt.Sprintf("tar -x -C %s -f -", baseDir)
 	plugin.DEBUG("Executing `%s`", cmd)
 	err = plugin.Exec(cmd, plugin.STDIN)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "@R{\u2717 Extract tar to temporary directory}\n")
+		fmt.Fprintf(os.Stderr, "@R{\u2717 Extract tar to temporry directory}\n")
```
patrick1378:
```suggestion
 		fmt.Fprintf(os.Stderr, "@R{\u2717 Extract tar to temporary directory}\n")
```
Quick spelling fix.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
 import (
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
-	fmt "github.com/jhunt/go-ansi"
+	fmt "github.com/jhunt/go-ansi"
```
jhunt: This whitespace is superfluous.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
-	DefaultPassword = "cassandra"
-	DefaultDataDir  = "/var/vcap/store/cassandra/data"
-
+	DefaultHost      = "127.0.0.1"
+	DefaultPort      = "9042"
+	DefaultUser      = "cassandra"
+	DefaultPassword  = "cassandra"
+	DefaultSaveUsers = true
+	DefaultDataDir   = "/var/vcap/store/cassandra/deployment_name/data"
+	DefaultBinDir    = "/var/vcap/jobs/cassandra/bin"
```
jhunt: Which BOSH release are these new defaults derived from?

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
+	DefaultExcludeKeyspaces = []string{"system_schema", "system_distributed", "system_auth", "system", "system_traces"}
+	SystemAuthTables        = []string{"roles", "role_permissions", "role_members", "resource_role_permissons_index"}
+)
+
 func main() {
 	p := CassandraPlugin{
 		Name:    "Cassandra Backup Plugin",
 		Author:  "Orange",
-		Version: "0.1.0",
+		Version: "1.0.0",
```
jhunt: Version bumps are not necessary; they are not actually communicated anywhere.  Plugin versions are locked to the version of SHIELD that is being distributed.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "host",
 				Type:     "string",
 				Title:    "Cassandra Host",
 				Help:     "The hostname or IP address of your Cassandra server.",
 				Default:  "127.0.0.1",
-				Required: true,
+				Required: false,
```
jhunt: If this is no longer required, we should remove the default.  If the default is standing in for "no value", then it should still be considered _required_.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "user",
 				Type:     "string",
 				Title:    "Cassandra Username",
 				Help:     "Username to authenticate to Cassandra as.",
 				Default:  "cassandra",
-				Required: true,
+				Required: false,
```
jhunt: If this is no longer required, we should remove the default.  If the default is standing in for "no value", then it should still be considered _required_.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
 			},
-			plugin.Field{
+			{
 				Mode:     "target",
-				Name:     "password",
+				Name:     "cassandra",
```
jhunt: Why are we changing the name of the password field from _password_ to _cassandra_?

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
 				Type:     "password",
 				Title:    "Cassandra Password",
-				Help:     "Password to authenticate to Cassandra as.",
+				Help:     "Password to auth_userenv",
```
jhunt: Elaborate on the use of this field.  What is `auth_userenv`, and how is it different from the prior text?

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
 				Default:  "cassandra",
-				Required: true,
+				Required: false,
```
jhunt: If this is no longer required, we should remove the default.  If the default is standing in for "no value", then it should still be considered _required_.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
 			},
-			plugin.Field{
+			{
 				Mode:     "target",
-				Name:     "keyspace",
-				Type:     "string",
-				Title:    "Keyspace to Backup",
-				Help:     "The name of the keyspace to backup.",
-				Example:  "system",
-				Required: true,
```
jhunt: For backwards-compatibility, we need to retain the `keyspace` field, but make it optional.  It should probably be mutually-exclusive with the new `include_keyspaces` / `exclude_keyspaces` options.  It would seem that a setting of `keyspace` should be interpreted as a setting of `include_keyspaces` containing a single element, the value.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
+				Name:     "include_keyspaces",
+				Type:     "array",
```
jhunt: Does this work in the SHIELD v8 UI?  I am uncertain that the UI can handle multi-valued fields and turn them into the proper JSON.  May need to revisit how keyspace is specified.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
+				Help:     "The name of the keyspace to exclude from backup.",
+				Example:  "[system']",
+				Default:  "['system_schema', 'system_distributed', 'system_auth', 'system', 'system_traces']",
+				Required: false,
+			},
+			{
+				Mode:     "target",
+				Name:     "save_users",
+				Type:     "boolean",
+				Title:    "to restore users and permissions along with the keyspaces that are restored",
```
jhunt: This will be rendered as the label on the v8 Web UI form for configuring this plugin, so it should probably be capitalized, and worded more appropriately.  Suggestion: "Also backup / restore users and permissions"

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
+				Help:     "The name of the keyspace to exclude from backup.",
```
jhunt: This help text is incorrect for this field.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
-				Name:     "keyspace",
-				Type:     "string",
-				Title:    "Keyspace to Backup",
-				Help:     "The name of the keyspace to backup.",
-				Example:  "system",
-				Required: true,
+				Name:     "include_keyspaces",
+				Type:     "array",
+				Title:    "Keyspaces to Include in the Backup",
+				Help:     "The name of the keyspace to include in the backup.",
```
jhunt: Since this is a list of multiple values, the help text should be pluralized.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
+				Example:  "['system']",
+				Required: false,
+			},
+			{
+				Mode:     "target",
+				Name:     "exclude_keyspaces",
+				Type:     "array",
+				Title:    "Keyspaces to Exclude from Backup",
+				Help:     "The name of the keyspace to exclude from backup.",
```
jhunt: Since this is a list of multiple values, the help text should be pluralized.

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
+				Example:  "true or false",
+				Required: false,
+			},
+			{
+				Mode:     "target",
+				Name:     "bindir",
+				Type:     "abspath",
+				Title:    "Path to binaries nodetool, sstableloader and cqlsh needed for the backup",
+				Help:     "where  to find binaries nodetool, sstableloader and cqlsh",
```
jhunt: The specific binaries needed should be listed in Help, not Title.  Title should be something suitable for an HTML form label, in the v8 Web UI, i.e. "Path to Cassandra binaries"

plugin/cassandra/plugin.go
```diff
@@ -90,106 +90,151 @@ package main
...
-			plugin.Field{
+			{
 				Mode:     "target",
 				Name:     "datadir",
 				Type:     "abspath",
 				Title:    "Path to Cassandra data/ directory",
 				Help:     "The absolute path to the data/ directory that contains the Cassandra database files.",
 				Default:  "/var/vcap/store/cassandra/data",
-				Required: true,
+				Required: false,
```
jhunt: If this is no longer required, we should remove the default.  If the default is standing in for "no value", then it should still be considered _required_.

plugin/cassandra/plugin.go
```diff
@@ -200,12 +245,15 @@ func main() {
...
-	DataDir  string
+	Host             string
+	Port             string
+	User             string
+	Password         string
+	IncludeKeyspaces []string
+	ExcludeKeyspaces []string
+	SaveUsers        bool
+	BinDir  string
+	DataDir string
```
jhunt: Looks like this code might need a `go fmt` before merge.

plugin/cassandra/plugin.go
```diff
@@ -285,6 +365,25 @@ func (p CassandraPlugin) Validate(endpoint plugin.ShieldEndpoint) error {
...
+		idx := sort.SearchStrings(excludeKeyspaces, keyspace)
+		if idx < len(excludeKeyspaces) && excludeKeyspaces[idx] == keyspace {
+			continue
+		}
+		savedKeyspaces = append(savedKeyspaces, keyspace)
+	}
+	sort.Strings(savedKeyspaces)
+
+	return savedKeyspaces
+}
```
jhunt: Can we remove some of the stutter in this code?  The work "Keyspace" appears often.  I believe `includeKeyspaces` could just be `include`, for example.

plugin/cassandra/plugin.go
```diff
@@ -324,53 +431,45 @@ func (p CassandraPlugin) Backup(endpoint plugin.ShieldEndpoint) error {
 	}
 	fmt.Fprintf(os.Stderr, "@G{\u2713 Create new snapshot}\n")
 
-	// Here we need to hard-link the snapshots/shield-backup directories into
-	// a {keyspace}/{tablename} structure that we'll temporarily put in
+	// Here we need to copy the snapshots/shield-backup directories into a
```
jhunt: I don't see evidence that we are copying now (instead of hard-linking).

plugin/cassandra/plugin.go
```diff
@@ -324,53 +431,45 @@ func (p CassandraPlugin) Backup(endpoint plugin.ShieldEndpoint) error {
...
-	err = plugin.Exec(cmd, plugin.STDOUT)
+	plugin.DEBUG("Creating base directories for '%s', with 0755 permissions", baseDir)
+	err = os.MkdirAll(baseDir, 0755)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "@R{\u2717 Clear base temporary directory}\n")
+		fmt.Fprintf(os.Stderr, "@R{\u2717 Create base temporary directory}\n")
 		return err
 	}
-	fmt.Fprintf(os.Stderr, "@G{\u2713 Clear base temporary directory}\n")
+	fmt.Fprintf(os.Stderr, "@G{\u2713 Create basemporary directory}\n")
```
jhunt: Typo "basemporary"; looks like a search-and-replace gone poorly 😁

plugin/cassandra/plugin.go
```diff
@@ -381,58 +480,63 @@ func (p CassandraPlugin) Backup(endpoint plugin.ShieldEndpoint) error {
 
 	info, err := os.Lstat(cassandra.DataDir)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "@R{\u2717 Recursive hard-link snapshot files}\n")
+		fmt.Fprintf(os.Stderr, "@R{\u2717 Recursive hard-link snapshot files in temp dir}\n")
```
jhunt: Can we spell out "temporary directory", everywhere we message it to the operator?

plugin/cassandra/plugin.go
```diff
@@ -485,67 +642,136 @@ func hardLinkAll(src_dir string, dst_dir string) (err error) {
...
-	err = os.MkdirAll("/var/vcap/store/shield/cassandra", 0755)
+	baseDir := "/var/vcap/store/shield/cassandra"
+
+	// Recursively remove /var/vcap/store/shield/cassandra, if any
+	cmd := fmt.Sprintf("rm -rf \"%s\"", baseDir)
+	plugin.DEBUG("Executing `%s`", cmd)
+	err = plugin.Exec(cmd, plugin.STDOUT)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "@R{\u2717 Create base temporary directory}\n")
+		fmt.Fprintf(os.Stderr, "@R{\u2717 Clean up any stale base temporary directory}\n")
```
jhunt: Either remove the word "any", or pluralize "directory".

plugin/cassandra/plugin.go
```diff
@@ -554,16 +780,39 @@ func (p CassandraPlugin) Restore(endpoint plugin.ShieldEndpoint) error {
...
 			return err
 		}
 	}
-	fmt.Fprintf(os.Stderr, "@G{\u2713 Load all tables data}\n")
+	return nil
+}
 
+func restoreUsers(cassandra *CassandraInfo, baseDir string) error {
+	plugin.DEBUG("Excluding cassandra user from 'system_auth.roles' table content")
+	cmd := fmt.Sprintf("sed -i -e '/^cassandra,/d' \"%s/system_auth.roles.csv\"", baseDir)
```
jhunt: This will fail if there is no `system_auth.roles.csv` file in the backup archive.  This causes issues restoring archives generated by current versions of this plugin, but it also proves problematic if an operator uses the new code, without saving users, and then decides to enable user-saving later.  All archives created before the switch will also break.

I believe we can just put a file check here (and elsewhere where new files are introduced) and behave accordingly.

OMI69-orange:
Hi everyone,
We think that we can close this PR, because we have made yesterday an addon plugin for shield agent , to not impact the SHIELD core version.
Thanks you for your remarks. I will integrate it in the next steps.

patrick1378:
We're still interested in accepting this as a contribution. We can assist with the changes and next steps, feel free to reach out.

Pat

gberche-orange:
Contrib mentionned by @OMI69-orange 

>we have made yesterday an addon plugin for shield agent 

is available at https://github.com/orange-cloudfoundry/shield-addon-cassandra-boshrelease

Issue #475, opened 2018-12-29
Agent test failing


jiangytcn:
$ go version                                                                                                                                                /home/jiangytcn 
go version go1.10.3 linux/amd64

~/go/src/github.com/starkandwayne/shield dvlp.yacloud.io on  master ⌚ 15:25:47
$ make all                                                                                                                                                  
go list ./... | grep -v vendor | xargs go fmt
plugin/webdav/plugin.go
go build ./cmd/shieldd
go build ./cmd/shield-agent
go build ./cmd/shield-schema
go build ./cmd/shield-crypt
go build ./cmd/shield-report
go build ./cmd/shield
go build ./cmd/shield-migrate
go build ./plugin/dummy
for plugin in $(cat plugins); do \
	go build ./plugin/$plugin; \
done
export PATH=$PATH:test/bin; go list ./... | grep -v vendor | xargs go test
Running Suite: Agent Test Suite

Random Seed: 1546068358
Will run 41 of 41 specs

• Failure [0.047 seconds]
Agent
/home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:14
  Command Runner
  /home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:264
    collects output from the command pipeline [It]
    /home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:352

    Expected
        <string>: {
          "compression": "bzip2",
          "key": "9ea61fef3024caadf35dd65d466a41fb51a3c152"
        }
        
    to match JSON of
        <string>: {
          "key": "9ea61fef3024caadf35dd65d466a41fb51a3c152"
        }

    /home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:363

• Failure [0.279 seconds]
Agent
/home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:14
  Command Runner
  /home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:264
    handles backup operations with large output [It]
    /home/jiangytcn/go/src/github.com/starkandwayne/shield/agent/agent_test.go:375

    Expected
        <string>: {
          "compression": "bzip2",
          "key": "acfd124b56584c471d7e03572fe62222ee4862e9"
        }
        
    to match JSON of
        <string>: {
          "key": "acfd124b56584c471d7e03572fe62222ee4862e9"
        }

[...]

PR #476, opened 2018-12-29
fix agent test issue

jiangytcn:
The default compression is `bzip2`, during the agent test, without providing compression parameter, leading to the failure during testing in #475 
https://github.com/starkandwayne/shield/blob/master/bin/shield-pipe#L110


jiangytcn:
seems the issue happened in my laptop.

PR #477, opened 2018-12-29
fix agent tests failue #475

jiangytcn:
using test's bin shield-report to output the execution result
address the issue of #475


jhunt:
Thanks!

PR #479, opened 2019-01-09
fix the issue with insecure s3-compabitle object storage server

jiangytcn:
for internal usage or testing purpose, the object storage service may be
configured in insecure mode, using the scheme specified in s3 host to
populate correct s3 api endpoint

### DIFF
--- a/plugin/s3/plugin.go
+++ b/plugin/s3/plugin.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


jhunt:
The code looks good, but I'd like to understand the issue you were facing that lead to this change needing to be made.  Can you elaborate further?

jiangytcn:
thanks James, 
I'm hosting a [minio](https://www.minio.io/) cluster as s3 compatible api for different system, shield is one of them as the backing up service,  for PoC and other purpose,  we didn't setup tls for object storage, which leading to the failure of s3 plugin to interact with. so I thought it would be helpful giving administrator sort of capability to control protocol of connection

PR #482, opened 2019-01-24
Add help for compression flag in CLI

thomasmitchell:
Secret's out. Now everybody knows you can use the compression flag with create-target and update-target.


jhunt:
@thomasmitchell can you incorporate review feedback into this PR?

Issue #483, opened 2019-01-25
Feature Add - support SSL connection for postgres plug-in


ChrisMcGowan:
Have the postgres plug-in support the option to enable SSL connection for backups and restores.

Recent releases of cf-deployment (https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L700) have TLS enabled for the uaa database as well as credhub as well.

I

PR #487, opened 2019-02-06
update postgres plugin to allow optional flags

### LINKED ISSUE #478: Enable custom attributes for backup and restore - Postgres plugin
Hi,

We would like to provide additional flags when taking a backup of a Postgres cluster, e.g _pg_dump -h <DB_IP> -Fc -o <DB>_. The current implementation of the plugin does not have the "options" parameter as the MySQL plugin.
As the backup won't be a plain SQL statement **psql** cannot be used to restore the database and a new option for **pg_restore** has to be used instead.

Is it possible to add these changes in the current plugin implementation? This will add more flexibility into the plugin allowing different use cases.

patrick1378:
This pr adds the functionality of adding optional flags to the pg_dump command of the postgres plugin. Additionally, I've added a change to the text of the cli help text for invite to reflect the correct roles.

This is to fix #478 and #483


cmd/shield/help.go
```diff
@@ -842,7 +842,7 @@ func ShowHelp(command string) {
 		fmt.Printf("  Local users (accounts defined wholly within SHIELD itself) can be\n")
 		fmt.Printf("  arbitrarily assigned to those tenants, and given one of three roles:\n")
 		fmt.Printf("\n")
-		fmt.Printf("    @M{manager}   Complete access to the tenant, including the ability\n")
+		fmt.Printf("    @M{admin}   Complete access to the tenant, including the ability\n")
```
jhunt: Is this related?

PR #512, opened 2019-05-08
Fix enum forms in web ui wizards

thomasmitchell:
The name field wasn't being populated, so it was appearing to not have a
value to the validation code. This meant any plugin using an enum for
its options could not be filled out on the web UI.


jhunt:
On second thought, we'll do release notes when we land some more changes that are in-flight.

PR #513, opened 2019-05-10
Add support for KV v2 in Vault plugin

thomasmitchell:
The code here... compiles. And its using the code we use for safe so it _probably_ works?

Need some testing done:
- Does it backup/restore a KV v1?
- Does it backup/restore a KV v2?
- Can it restore a backup made with a previous version of this plugin?
- Can it restore a v1 to a v2 backend? (it should be able to)
- How does it behave when trying to restore a v2 backend with multiple versions into a v1 mount? (and how do we want it to behave?)


jhunt:
CI tests are failing in Travis.  Please fix.

jhunt:
The transient HTTP 502 is gone, but now we have issues since it doesn't seem like the vaultkv lib got embedded into `vendor/`; can you take a look @thomasmitchell ?

thomasmitchell:
Vendoring was done.

jhunt:
This looks like it works; i'm going to hold off on approving though, until we can get validation that the code works on v2 (manual tests) and that v1 backups from previous versions of the plugin can be restored with the proposed changes.

jhunt:
Now that Phalanx is working, we have automated tests for Vault in a purely-v2 environment.  Merging this.

PR #528, opened 2019-05-23
Add support for Openstack V3 identity API

peterellisjones:
Hi Shield team,

The [Openstack golang client](https://github.com/openstack/golang-client) that was previously being used does not support the [Openstack V3 identity API](https://developer.openstack.org/api-ref/identity/v3/). This commit updates the Swift plugin to use [github.com/ncw/swift](https://github.com/ncw/swift) Swift client which supports V3 as well as V2. 

I have verified that these changes allow the plugin to backup and restore data using both the V3 and V2 API, using [Redis dedicated nodes](https://github.com/pivotal-cf/cf-redis-release) as a data source.

Notes:
* Using the V3 API requires a new field `domain`. I've updated the docs to reflect this
* The new swift client doesn't have a debug mode which prints the HTTP requests, so I removed that property and updated the docs to reflect it.

Manual test I did for checking that the changes works (for both V2 and V3 APIs):
1. Move compiled plugin into /var/vcap/packages/shield/plugins/swift
2. Make a backup (backup 1)
3. Use Redis CLI to write data
4. Make a backup (backup 2)
5. Restore backup 1 
6. Verify data is not present
7. Restore backup 2
8. Verify data is present

kind regards,

Pete


jhunt:
🎉 Awesome!  Thank you so much for taking the time to do this!

This looks great to me, and passes all CI, so I'm going to merge it post-haste.

PR #573, opened 2019-07-26
etcd plugin implemented with shield

sriniketh923:


plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+			"endpoints" : "https://192.168.42.45:2379"   # REQUIRED
+
+			"auth"  : false												# is role based or cert based auth enabled on the etcd cluster
+			"user"  : "admin",            								# username for role based authentication
+			"pass"  : "p@ssw0rd"          								# password for role based authentication
+			"clientCertPath" : "/tmp/test-certs/test-client.crt" 		# path to client certificate
+			"clientKeyPath"  : "/tmp/test-certs/test-client-key.key" 	# path to client key
+			"caCertPath"     : "/tmp/test-certs/test-ca.crt" 			# path to CA certificate
+			"fullOverwrite"  : "false"								    # enable or disable full overwrite of the cluster
+			"prefix"		 : "starkandwayne/"							# backup specific keys
```
jhunt: Check the help output of the plugin, and adjust white space as necessary to make it more aesthetically pleasing.  Replace tabs with spaces while you're at it.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+			{
+			}
+			`,
+		Fields: []plugin.Field{
+			plugin.Field{
+				Mode:     "target",
+				Name:     "endpoints",
+				Type:     "string",
+				Title:    "Endpoint",
+				Help:     "IP addresses of the etcd nodes in the cluster",
```
jhunt: These are more than just IP addresses, and the help should reflect that.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+		Defaults: `
+			{
+			}
+			`,
+		Fields: []plugin.Field{
+			plugin.Field{
+				Mode:     "target",
+				Name:     "endpoints",
+				Type:     "string",
+				Title:    "Endpoint",
```
jhunt: I would rename this field to emphasize that it is the upstream etcd client endpoint, and not the peer URLs.  Maybe `etcd URLs`?

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Title:   "Password",
+				Example: "p@ssw0rd",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "clientCertPath",
+				Type:    "string",
+				Help:    "Path to the certificate issued by the CA for the client connecting to the ETCD cluster",
+				Title:   "Client Certificate File Path",
+				Example: "/tmp/test-certs/test-client.crt",
```
jhunt: Don't rely on paths; let the operator specify the full PEM contents of the client certificate.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Mode:    "target",
+				Name:    "pass",
+				Type:    "password",
+				Help:    "Password for role based authentication",
+				Title:   "Password",
+				Example: "p@ssw0rd",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "clientCertPath",
```
jhunt: Field names should be snake_case not camelCase.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Title:   "Client Key File Path",
+				Example: "/tmp/test-certs/test-client-key.key",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "caCertPath",
+				Type:    "string",
+				Help:    "Path to the CA certificate that issued the client cert and key",
+				Title:   "Trusted CA File Path",
+				Example: "/tmp/test-certs/test-ca.crt",
```
jhunt: Don't rely on paths; let the operator specify the full PEM contents of the CA certificate.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Title:   "Client Certificate File Path",
+				Example: "/tmp/test-certs/test-client.crt",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "clientKeyPath",
+				Type:    "string",
+				Help:    "Path to the key issued by the CA for the client connecting to the ETCD cluster",
+				Title:   "Client Key File Path",
+				Example: "/tmp/test-certs/test-client-key.key",
```
jhunt: Don't rely on paths; let the operator specify the full PEM contents of the client key.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Mode:    "target",
+				Name:    "caCertPath",
+				Type:    "string",
+				Help:    "Path to the CA certificate that issued the client cert and key",
+				Title:   "Trusted CA File Path",
+				Example: "/tmp/test-certs/test-ca.crt",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "fullOverwrite",
```
jhunt: Rename this field to just `overwrite`

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Type:    "bool",
+				Help:    "If this is enabled, the existing key/value pairs will be deleted. The values will be restored using the backup archive.",
```
jhunt: It's worth pointing out here that only keys underneath `prefix` will be deleted.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Title:   "Full Overwrite",
+				Example: "false",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "prefix",
+				Type:    "string",
+				Help:    "This is the input string for prefix based backup-restore",
+				Title:   "Prefix",
+				Example: "starkandwayne/",
```
jhunt: If the trailing backslash is necessary that should be stated.  If the plugin can supply it, that's even better.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Example:  "https://192.168.42.45:2379",
+				Required: true,
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "auth",
+				Type:    "bool",
+				Title:   "Authentication",
+				Help:    "Is role based or cert based authentication enabled on the etcd cluster",
+				Example: "false",
```
jhunt: Is it possible to have no authentication on etcd?

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Mode:    "target",
+				Name:    "user",
+				Type:    "string",
+				Help:    "Username for role based authentication",
+				Title:   "Username",
+				Example: "admin",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "pass",
```
jhunt: Use the full word, `password` for the field name.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+				Mode:    "target",
+				Name:    "auth",
+				Type:    "bool",
+				Title:   "Authentication",
+				Help:    "Is role based or cert based authentication enabled on the etcd cluster",
+				Example: "false",
+			},
+			plugin.Field{
+				Mode:    "target",
+				Name:    "user",
```
jhunt: Use the full word, `username` for the field name.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	auth, err := endpoint.BooleanValueDefault("auth", false)
+	if err != nil {
+		return nil, err
+	}
+
+	user, err := endpoint.StringValueDefault("user", "")
```
jhunt: If authentication is enabled, and a username / password is required, you should enforce that, and not use `StringValueDefault()` in that case.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	pass, err := endpoint.StringValueDefault("pass", "")
+	if err != nil {
+		return nil, err
+	}
+
+	clientCert, err := endpoint.StringValueDefault("clientCertPath", "")
```
jhunt: If X.509 certificates are required, based on auth mode, you should ensure that they are valid here, and not use `StringValueDefault()`.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+
+	fullOverwrite, err := endpoint.BooleanValueDefault("fullOverwrite", false)
+	if err != nil {
+		return nil, err
+	}
+
+	prefix, err := endpoint.StringValueDefault("prefix", "")
+	if err != nil {
+		return nil, err
+	}
```
jhunt: If prefix needs to be fixed up (i.e. a trailing slash added) do so here.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+		Password:       pass,
+		ClientCertPath: clientCert,
+		ClientKeyPath:  clientKey,
+		CaCertPath:     caCert,
+		FullOverwrite:  fullOverwrite,
+		Prefix:         prefix,
+	}, nil
+}
+
+func (p EtcdPlugin) Validate(endpoint plugin.ShieldEndpoint) error {
```
jhunt: The output of this function needs to be firmed up and made more legible.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+		fmt.Printf("@G{\u2713 auth} authentication is disabled\n")
+	}
+
+	if b {
+		s, err = endpoint.StringValueDefault("user", "")
+		if err != nil {
+			fmt.Printf("@R{\u2717 user  %s}\n", err)
+			fail = true
+		} else if s == "" {
+			fmt.Printf("@G{\u2713 user} username was not provided so cert based auth will be used\n")
```
jhunt: This is probably a bad idea; change `auth` to a string / enum, and be more explicit about value management.  I.e. if auth is `x509`, do cert stuff, if it's `none`, don't do auth, etc.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	b, err = endpoint.BooleanValueDefault("fullOverwrite", false)
+	if err != nil {
+		fmt.Printf("@R{\u2717 full restore  %s}\n", err)
+		fail = true
+	} else if b {
+		fmt.Printf("@G{\u2713} full restore enabled\n")
+	} else {
+		fmt.Printf("@G{\u2713} full restore disabled\n")
```
jhunt: This verbiage is a bit confusing to someone not steeped in how the plugin works.  Consider rewording to explain what overwrite means, and what the scope is.  It's okay to use the value from `prefix` in the messaging here (assuming you validate it beforehand)

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	cli, err := clientv3.New(clientv3.Config{
+		Endpoints:   []string{etcd.EtcdEndpoints},
+		DialTimeout: 2 * time.Second,
+		Username:    etcd.Username,
+		Password:    etcd.Password,
+		TLS:         tlsConfig,
+	})
```
jhunt: Most of the preceding code can be moved into a method on `*EtcdConfig`, and eliminate code duplication.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+
+	if err != nil {
+		log.Fatal(err)
```
jhunt: Just return the error and let the SHIELD plugin framework handle exiting and logging.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+			}
+			defer cancel()
+		} else {
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+			_, err := cli.Delete(ctx, etcd.Prefix, clientv3.WithPrefix())
+			if err != nil {
+				log.Fatal(err)
+			}
+			defer cancel()
+		}
```
jhunt: The consequent and alternate clauses of the inner if/else block are identical, given the invariant established by the condition.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+
+		dataval, err := base64.StdEncoding.DecodeString(strings.Split(string(lineBuffer), " : ")[1])
+		if err != nil {
+			fmt.Printf("error decoding value: %s", err)
+			return err
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		_, err = cli.Put(ctx, fmt.Sprintf("%s", datakey), fmt.Sprintf("%s", dataval))
+		defer cancel()
```
jhunt: Is deferring cancel() in a for loop wise?

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,451 @@
...
+		if err != nil {
+			log.Fatal(err)
```
jhunt: Return the error, don't log.Fatal it.

plugin/etcd/plugin.go
jhunt: Y'all must be windows users.  😛 

I would try stripping of all trailing `/` characters and then appending a single `/`.  The standard `strings` library can be quite helpful in this.

plugin/etcd/plugin.go
jhunt: Change this to just `urls`

plugin/etcd/plugin.go
```diff
@@ -46,10 +45,10 @@ func main() {
 		Fields: []plugin.Field{
 			plugin.Field{
 				Mode:     "target",
-				Name:     "endpoints",
+				Name:     "etcd_url",
 				Type:     "string",
 				Title:    "Endpoint",
-				Help:     "IP addresses of the etcd nodes in the cluster",
+				Help:     "It is the upstream etcd client endpoint",
```
jhunt: Change this to "The client URL(s) of the etcd cluster.  Separate multiple URLs with commas."

plugin/etcd/plugin.go
sriniketh923: The trailing backslash was not necessary.

plugin/etcd/plugin.go
sriniketh923: The trailing backslash was not necessary.

plugin/etcd/plugin.go
```diff
@@ -45,76 +49,90 @@ func main() {
...
 			},
 			plugin.Field{
-				Mode:    "target",
-				Name:    "auth",
-				Type:    "bool",
+				Mode:     "target",
+				Name:     "timeout",
+				Type:     "string",
+				Title:    "Dial Timeout",
+				Help:     "DialTimeout is the timeout for failing to establish a connection. Enter time in seconds.",
```
jhunt: Dial is a decidedly "Go" thing to say (thanks AT&T) - I would retitle as 'Connection Timeout' or just 'Timeout' and elude to the fact that it is timing out connections in the help text.

plugin/etcd/plugin.go
```diff
@@ -45,76 +49,90 @@ func main() {
...
+				Example:  "9",
+				Required: true,
```
jhunt: We can go ahead and default this to 2.

plugin/etcd/plugin.go
```diff
@@ -27,15 +29,17 @@ func main() {
 		},
 		Example: `
 			{
-			"etcd_url"   : "https://192.168.42.45:2379"                                                                        # REQUIRED
-			"auth"        : false                                                                                               # is role based or cert based auth enabled on the etcd cluster
+			"url"         : "https://192.168.42.45:2379,https://192.168.23.54:2379"                                             # REQUIRED
+            "timeout"     : "9"                                                                                                 # REQUIRED
```
jhunt: Errant white space issue.

plugin/etcd/plugin.go
```diff
@@ -125,28 +143,37 @@ func main() {
...
 func getEtcdConfig(endpoint plugin.ShieldEndpoint) (*EtcdConfig, error) {
-	etcdEndpoint, err := endpoint.StringValue("etcd_url")
+	etcdEndpoint, err := endpoint.StringValue("url")
+	if err != nil {
+		return nil, err
+	}
+
+	etcdUrls := strings.Split(etcdEndpoint, ",")
+
+	timeoutStr, err := endpoint.StringValue("timeout")
```
jhunt: Default to 2 seconds.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,496 @@
...
+            "password"    : "p@ssw0rd"                                                                                          # password for role based authentication
+            "client_cert" : "-----BEGIN CERTIFICATE-----\n(cert contents)\n(... etc ...)\n-----END CERTIFICATE-----"            # path to client certificate
+            "client_key"  : "-----BEGIN RSA PRIVATE KEY-----\n(cert contents)\n(... etc ...)\n-----END RSA PRIVATE KEY-----"    # path to client key
+            "ca_cert"     : "-----BEGIN CERTIFICATE-----\n(cert contents)\n(... etc ...)\n-----END CERTIFICATE-----"            # path to CA certificate
+            "overwrite"   : "false"                                                                                             # enable or disable full overwrite of the cluster
+            "prefix"      : "starkandwayne"                                                                                     # backup specific keys
+            }
+            `,
+		Defaults: `
+            {
```
jhunt:
```suggestion
            {
              "timeout": "2"
          
```

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,496 @@
...
+				Example: "2",
+			},
+			plugin.Field{
+				Mode: "target",
+				Name: "auth",
+				Type: "enum",
+				Enum: []string{
+					"Role-Based Authentication",
+					"Certificate-Based Authentication",
+					"Both",
```
jhunt: Use your shiny new variables here too.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,496 @@
...
+		return nil, err
+	}
+
+	prefix, err := endpoint.StringValueDefault("prefix", "")
+	if err != nil {
+		return nil, err
+	}
+
+	tlsConfig := &tls.Config{}
+	if auth == "Certificate-Based Authentication" {
```
jhunt: Use those shiny new variables here as well.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,496 @@
...
+		caCertPool := x509.NewCertPool()
+		check := caCertPool.AppendCertsFromPEM([]byte(caCert))
+		if check != true {
+			plugin.DEBUG("CA cert did't parse right.\n")
+		}
```
jhunt: This should be set up if `ca_cert` is set to _anything_, in case I want to run etcd with my own internal CA-issued certificate, but do _not_ want to use anything other than role-based authentication (or don't auth at all!)

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,496 @@
...
+	var (
+		s    string
+		b    bool
+		err  error
+		fail bool
+	)
+
+	roleAuth := "Role-Based Authentication"
+	certAuth := "Certificate-Based Authentication"
+	bothAuth := "Both"
```
jhunt: Make these global variables and use them everywhere.  Put them at the top of the file, and name them something like `RoleBasedAuthentication`, `CertificateAuthentication`, etc.

plugin/etcd/plugin.go
```diff
@@ -0,0 +1,496 @@
...
+			fail = true
+		} else if s == "" {
+			fmt.Printf("@R{\u2717 client_key}            Certificate based authentication chosen but client private key was not provided\n")
+			fail = true
+		} else {
+			/* FIXME: validate that it is an X.509 PEM certificate */
+			fmt.Printf("@G{\u2713 client_key}			 Key present\n")
+		}
+
+		s, err = endpoint.StringValue("ca_cert")
```
jhunt: Run this validation regardless of authentication type.

PR #576, opened 2019-08-07
Etcd plugin

sriniketh923:
Fixed bug in CA certificate validation


jhunt:
@sriniketh923 can you rebase this against new origin/master?  I did some bad things to HEAD that affected you (inadvertently)

jhunt:
I went ahead and made the fixes directly.  Good catch though!

PR #580, opened 2019-08-12
Added help notes about property name changes

thomasmitchell:
This changed in 8.0.17 and broke backward compatibility. Worth noting that it happened in the help.


jhunt:
These will show up in the web UI (where they are not particularly useful).  I think we'd be better off either documenting them on the website.

Issue #460, opened 2018-10-08
"Every-x-hours" sets "0 backups for x days"


jordanbcooper:
When setting "every x hours" (in my case, via import) backup jobs specify "keep 0 backups (x days)"

defhand:
@jhunt Any reccomendations here?

jhunt:
I just ran some tests in our lab, running 8.0.17, and here's what I found.

1. The timespec `every 4 hours` is a syntax error.
2. The timespec `every 4 hours at 12:15am` is also a syntax error.
3. The timespec `every 4 hours from 12:15am` is _not_ a syntax error.

When you configure a backup job via the web UI, the every-x-hours flavor of scheduling comes out in text(timespec) as `every X hours from HH:MMam`

Here's a screenshot!

<img width="819" alt="Every X Hours (bottom schedule)" src="https://user-images.githubusercontent.com/20047/54370286-bbe0aa80-464d-11e9-9c35-4c483c755f64.png">

I'm curious if we are dealing with an edge case I just can't find, or if the import process is incorrectly parsing, or just not validating the input schedule timespec.

@jordanbcooper  / @defhand what timespec strings are you sending up to SHIELD to evoke this behavior?

jordanbcooper:
Looks like i'm sending 

`    jobs:
    - name:  job-name
      policy: Long-Term
      storage: S3 - database
      when: every 3 hours from 12:00AM`

jordanbcooper:
Er, that didn't fix it. My polices are Short-Term and Long-Term and are applied to each job accordingly. Strangely, my "daily 4am" job says "keep 90 backups (90 days)"

jhunt:
Previously, on _"Every-x-hours" sets "0 backups for x days"_:

I originally thought we were dealing with bad validation on the import command / API handler, so I tried this import file:

```
core: http://127.0.0.1:8180
token: 353425ce-ccfe-4e10-8439-f3dce810dc48

tenants:
  - name: Acme, Inc.
    systems:
      - name:    'Jordan (at)'
        summary: A test system for generating lots of web UI output and pushing the limits of the web UI
        agent:   vault:5444
        plugin:  fs
        config:
          base_dir: /tmp
        jobs:
          - name:     Fast AT
            when:     every 3 hours at 12:05am
            paused:   yes
            storage:  S3
            policy:   Long-Term
```

when I ran that, however, I got this:

```
Connecting to http://127.0.0.1:8180
Importing Users...

[...]

PR #590, opened 2019-08-22
Name the #460 fixup in code

jhunt:


jhunt:
Withdrawing this; future change (forthcoming) to the fixups code will conflict, so I'll resubmit.

PR #593, opened 2019-08-22
More Rigorous Fixup Handling

jhunt:
We now only run the fixups once (although the code does allow re-runs,
should we ever plumb in an API / UI for this).

### DIFF
--- a/db/fixup.go
+++ b/db/fixup.go
@@ -179,7 +179,7 @@ See [issue #460](https://github.com/shieldproject/shield/issues/460) in GitHub f
 				job.KeepN = sched.KeepN(job.KeepDays)
 				err = db.doUpdateJob(job)
 				if err != nil {
-					/* FIXME log stuff! */
+					log.Errorf("failed to apply UpdateJob fixup: %s", err)
 					continue
 				}
 			}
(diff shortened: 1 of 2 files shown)


db/fixup.go
```diff
@@ -128,79 +173,86 @@ See [issue #460](https://github.com/shieldproject/shield/issues/460) in GitHub f
 
 				sched, err := timespec.Parse(job.Schedule)
 				if err != nil {
-					log.Errorf("failed to parse job schedule '%s': %s", job.Schedule, err)
+					/* FIXME log stuff! */
```
patjones:
```suggestion
					log.Errorf("failed to parse job schedule '%s': %s", job.Schedule, err)
```

db/fixup.go
```diff
@@ -128,79 +173,86 @@ See [issue #460](https://github.com/shieldproject/shield/issues/460) in GitHub f
...
 					continue
 				}
 				job.KeepN = sched.KeepN(job.KeepDays)
 				err = db.doUpdateJob(job)
 				if err != nil {
-					log.Errorf("failed to apply UpdateJob fixup: %s", err)
+					/* FIXME log stuff! */
```
patjones:
```suggestion
					log.Errorf("failed to apply UpdateJob fixup: %s", err)
```

db/fixup.go
patjones: Adding back in improved error checking.

PR #598, opened 2019-08-23
Expert SQL API

### LINKED ISSUE #519: [NEW] "Expert Mode" SQL
**Is your feature request related to a problem? Please describe.**
We've been trying to validate things with direct database queries in testdev and lab environments, and we regularly get `database is locked` because of the high contention for db access.

**Describe the solution you'd like**
I would like to have an "expert mode" in the admin backend.  This should be scary.  Lots of red.  LOTS OF RED.  In this expert mode, I would like to have a SQL query (SELECT only) mechanism that lets operators explore the database.  This should only be available to administrators.

**Describe alternatives you've considered**
Direct database access is the only alternative, and it's a bit annoying.  This solution also allows us to troubleshoot end-user sites, by sending them queries to run and getting results back, without direct box access.

jhunt:
The `/v2/expert/sql` API and `shield op sql` command now allow SHIELD
oeprators to execute arbitrary (and I mean ARBITRARY) SQL against the
SHIELD Core database.

Fixes #519

### DIFF
--- a/cmd/shield/main.go
+++ b/cmd/shield/main.go
@@ -327,6 +327,7 @@ var opts struct {
 	Op struct {
 		Pry struct{} `cli:"pry"`
 		IFK struct{} `cli:"ifk"`
+		SQL struct{} `cli:"sql"`
 	} `cli:"op"`
 }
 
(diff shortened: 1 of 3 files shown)


cmd/shield/main.go
```diff
@@ -3286,6 +3287,57 @@ tenants:
...
+			fmt.Fprintf(os.Stderr, "@R{error: %s}\n", out.Error)
+			if out.Diagnostic != "" {
+				fmt.Fprintf(os.Stderr, "@R{       %s}\n", out.Diagnostic)
+			}
+			os.Exit(1)
+		}
+
+		var out struct {
+			Columns []string `json:"columns"`
+			Rows []map[string] interface{} `json:"rows"`
```
thomasmitchell: Fix that space or somebody else's gofmt is going to get it later.

jhunt: go fmt'd ftw

PR #599, opened 2019-08-24
WebSocket Improvements

### LINKED ISSUE #541: [NEW] Stats on Websocket Usage / Traffic
**Is your feature request related to a problem? Please describe.**
I've noticed some strange slowdowns with SHIELD that seem to be related to leakage in the Web Socket.  I have no basis of fact for this theory, it just feels like the Web Socket is to blame.

I'd like to get some insight into the health of the Web Socket / messagebus.

**Describe the solution you'd like**

Ideally, there would be a (dev-only?) endpoint for retrieving statistics from SHIELD about mbus / websocket usage.  Those metrics would include:

- Total number of connected clients (ever)
- Current number of connected clients
- Total number of messages, of each type, sent (ever)
- Total size of messages, of each type, sent (ever)

**Describe alternatives you've considered**
We've tried to use stock Go debugging to get things like profile traces and other stuff, but it doesn't seem to show anything unusual.

jhunt:
- We no longer leak file descriptors / goroutines for websocket clients
  who have gone away.

- We now get more metrics from `/v2/mbus/status`, regarding connections,
  message counts and payloads, etc.

I think this fixes some bizarre networking deadlock we saw on long-lived
(and much loved) SHIELD instances, in our lab.

Fixes #541

### DIFF
--- a/core/api_v2.go
+++ b/core/api_v2.go
@@ -505,20 +505,31 @@ func (c *Core) v2API() *route.Router {
 		}
 
 		log.Infof("registering message bus web client")
-		ch, _, err := c.bus.Register(queues)
+		ch, slot, err := c.bus.Register(queues)
 		if err != nil {
 			r.Fail(route.Oops(err, "Unable to begin streaming SHIELD events"))
 			return
 		}
+		log.Infof("registered with message bus as [slot:%d]", slot)
 
-		go socket.Discard()
+		go socket.Discard(func () { close(ch) })
 		for event := range ch {
 			b, err := json.Marshal(event)
 			if err != nil {
-				log.Errorf("message bus web client failed to marshal JSON for websocket relay: %s", err)
+				log.Errorf("message bus web client [slot:%d] failed to marshal JSON for websocket relay: %s", slot, err)
 			} else {
-				socket.Write(b)
+				if done, err := socket.Write(b); done {
+					break
+				} else if err != nil {
+					log.Errorf("failed to write message to message bus web client [slot:%d]: %s", slot, err)
+				}
 			}
+		}
+
+		log.Infof("message bus web client [slot:%d] disconnected; unregistering...", slot)
+		err = c.bus.Unregister(slot)
+		if err != nil {
+			log.Errorf("message bus web client [slot:%d] failed to unregister after disconnect: %s", slot, err)
 		}
 	})
 	// }}}
--- a/core/bus/bus.go
+++ b/core/bus/bus.go
@@ -27,6 +27,12 @@ type Event struct {
 type Bus struct {
 	lock  sync.Mutex
 	slots []slot
+
+	lifetime, current struct {
+		connections int64
+	}
+	events   map[string]int64
+	messages map[string]int64
 }
 
 type slot struct {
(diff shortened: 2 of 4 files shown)


core/api_v2.go
```diff
@@ -505,21 +505,35 @@ func (c *Core) v2API() *route.Router {
...
 		if err != nil {
 			r.Fail(route.Oops(err, "Unable to begin streaming SHIELD events"))
 			return
 		}
+		log.Infof("registered with message bus as [slot:%d]", slot)
 
-		go socket.Discard()
+		go socket.Discard(func() {
+			defer recover()
+			close(ch)
```
thomasmitchell: closing this channel here is redundant, as it is also closed in Unregister. Closing a closed channel will result in a panic.

thomasmitchell: However, it is necessary to break out of the loop. We probably need to do a non-blocking read in Unregister to determine if the channel has already been closed.

core/bus/bus.go
```diff
@@ -95,6 +106,14 @@ func (b *Bus) SendEvent(queues []string, ev Event) {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
+	if _, ok := b.events[ev.Event]; !ok {
+		b.events[ev.Event] = 0
+	}
+	if _, ok := b.messages[ev.Event]; !ok {
+		b.messages[ev.Event] = 0
+	}
```
thomasmitchell: These initializations are actually unnecessary, as referencing a non-existent map entry returns its zero value. `b.events[iDontExist]++` would set the value to 1 without issue. It's up to you if you want to keep the initialization though. It's not hurting anything.

core/bus/bus.go
```diff
@@ -67,12 +78,12 @@ func (b *Bus) Unregister(idx int) error {
 		return fmt.Errorf("could not unregister channel #%d: index out of range", idx)
 	}
 
+	b.current.connections -= 1
```
thomasmitchell: Unregister should probably set some timer to delete the events / messages count from the metrics map, or else we technically have a small memory leak. I doubt that the system would ever get overtaken by an army of int64s, but... up to you how much you care about that.

core/bus/bus.go
thomasmitchell: I see the purpose now in the DumpState loop, so scratch this.

core/bus/bus.go
jhunt: Event types are a finite set; the main problem we might run into is overflow, but the idea of monitoring this is to use something else (like RRD) to detect rollover and adjust the delta calculation. In that respect, the messages/bytes counters work more like packet/byte counters in network gear — ever-increasing.

core/api_v2.go
jhunt: I found the “ignore panic” of the `defer recover()` works well.

thomasmitchell: Yes, but this isn't the thread that is going to panic. This will fire first if the client closes the websocket, and then Unregister will call it the second time and then panic.

core/bus/bus.go
thomasmitchell: Missed that this is keying on the event type string and not the event object itself - which makes sense because you can't use a struct as a key.

thomasmitchell: If overflow is a real concern, the `big` package is a thing.

core/api_v2.go
jhunt: Other option might be to `close(ch)` in the `done` branch of `socket.Write(b)` (new file line 524), right before we break out of the loop, and then _not_ close the channel in Unregister;  I'm unclear if anyone else is calling Unregister out of band or not.

jhunt: Belt, meet suspenders. bc9f73a has a fix that is cautiously safe (I think).

core/bus/bus.go
jhunt: I'm really not.

PR #600, opened 2019-08-28
Mbus queues are buffered; sending events is async

thomasmitchell:
Mbus channels have a buffer size - if a message is attempted to be put
in a full queue, the channel will be closed and the slot deregistered.
The buffer size is configurable as "slot-size". Additionally, the number
of slots that are initialized is now configurable as "max-slots".

Mbus event sends are asynchronous now. Calls to SendEvent should now
return in a predictable amount of time instead of potentially needing to
wait for a message to be read from a channel synchronisly.

The mbus/status endpoint has been altered: it only lists the slots that
are active at that moment (to avoid dumping 2047 empty slots, for
example). Each slot reports its current channel len and high watermark
channel length now. The status also reports the configuration of the
mbus as a whole.

Register now returns a unique ID instead of the slot index. This allows
calls to Unregister to be idempotent, which allows users of the mbus to
more realistically allow the mbus to handle all cases of the closing of
the message channel (through Unregister). Because Unregister can be
synchronized with other channel accesses (because its owned by the Mbus
object), we eliminate the possibility of send/close races, leading to
less requirements to recover from panics.

The API now attempts to be nice and send a close message to the client
in the case of a write error, giving the client a suggestion to
dis/reconnect. Without this, the client may be unaware of the write
error, and may keepalive the connection with the belief that more
messages are en-route someday. The close message can timeout on send -
it is only a best effort attempt to guide the client towards the correct
behavior.

Websocket message writes now have a timeout. It is currently hardcoded
to 45 seconds, but someday can be made configurable if necessary. In the
case of a write timeout, the queue is Unregistered and a close message
is attempted to be sent to the client.

The Write function now properly reports when a websocket has been closed.

Additional work likely needs to be done on the web UI side to make it
better at reconnecting in the case of a closed websocket, but these
changes should cause a websocket to be closed when conditions become
bad, and also allow SHIELD to continue working normally in the case of a
single bad websocket client.

### DIFF
--- a/core/api_v2.go
+++ b/core/api_v2.go
@@ -504,42 +504,45 @@ func (c *Core) v2API() *route.Router {
 			return
 		}
 
+		socket.SetWriteTimeout(45 * time.Second)
+
 		log.Infof("registering message bus web client")
 		ch, slot, err := c.bus.Register(queues)
 		if err != nil {
 			r.Fail(route.Oops(err, "Unable to begin streaming SHIELD events"))
 			return
 		}
-		log.Infof("registered with message bus as [slot:%d]", slot)
+		log.Infof("registered with message bus as [id:%d]", slot)
 
-		closeMeSoftly := func () {
-			defer recover()
-			close(ch)
-		}
+		closeMeSoftly := func() { c.bus.Unregister(slot) }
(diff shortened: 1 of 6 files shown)


jhunt:
LGTM!

PR #604, opened 2019-09-04
Differential backup feature

ssaadawi-orange:
Here is a pull request for differential backup & restore. The feature is retro-compatible with former version and implemented following the xtrabackup specs : https://www.percona.com/doc/percona-xtrabackup/2.4/backup_scenarios/incremental_backup.html.


ssaadawi-orange:
Hi James, 

You are totally correct, we implemented the solution that way as it was convenient enough for us. Indeed, the feature lacks genericity and automation in its behaviour. We'll figure if it's something doable for the time allocated for it on our end to bring up something more in the standards and I'll update this PR in accordance. Have a great day.

Sami.

PR #611, opened 2019-09-10
Fix Websocket timeouts

thomasmitchell:
The server now sends keepalive messages after a configurable timeout
interval if a message has not been sent during that time. This should
keep the connection open even if there are no events to send.

The write timeout for messages is now configurable.

The configuration will now error if mbus parameters are negative.


thomasmitchell:
@sriniketh923 This is the branch that I could use some help with testing :)

sriniketh923:
Got it. I'll take a look at it

PR #640, opened 2019-10-25
New mbus health event

sriniketh923:
Implemented new health status messages for targets, stores and jobs


jhunt:
@sriniketh923 If we're putting the health state in the database (a move I'm quite happy with) do we really need a health-update event?  Or is health-update just a special case of update, and we dispatch in the front-end to determine when / how to update things like HUD?

sriniketh923:
@jhunt Yes health-update is a special case of the update. On the front-end side the health-update event updates the object just like update-object event. Having a separate health-update event might make it easier for operators to keep track of the health should something go wrong, but not having this event will not effect front-end for updating things like the HUD.

PR #644, opened 2019-11-12
Daily Storage Increase can use larger units

thomasmitchell:
Previously, if the daily storage increase in the web UI was negative,
then it would always display the amount in bytes, no matter how large
the number becomes. This should allow the unit formatting to work even
when the number is negative.


thomasmitchell:
I have not tested this as of yet, just throwing the simple code out there for now. If anybody has a chance to run it through a test before I get back around to it, I would appreciate it.

PR #651, opened 2019-11-14
Case insensitive web UI sorting

### LINKED ISSUE #646: [BUG] Sorting in Web UI is case-sensitive
**Describe the bug**

If I sort the following alphabetically
```
Shield
animaniacs
```

...I expect animaniacs to be first in the list. In the Web UI, this is not the case, because it
just uses ASCII value ordering. This can be fixed by lowercase-ing all input before sorting.

**To Reproduce**
1. Make two jobs, one with a capital letter at the start and one without.
2. Sort by name
3. Notice the bad sorting.

**Expected behavior**
Case insensitive sorting

**Screenshots**

**SHIELD versions (please complete the following information):**
 - SHIELD Core: 8.5.0
 - SHIELD CLI: n/a
**Browser version(s) (please complete if reporting a web UI bug):**
 - OS: OS X
 - Browser: Firefox
 - Version: 70

**Additional context**

sriniketh923:
Fixes #646

### DIFF
--- a/web/htdocs/js/events.js
+++ b/web/htdocs/js/events.js
@@ -131,8 +131,8 @@
         });
 
         rows.sort(function (a, b) {
-          return mode * (a[0] > b[0] ?  1 :
-                         a[0] < b[0] ? -1 : 0);
+          return mode * (a[0].toLowerCase() > b[0].toLowerCase() ?  1 :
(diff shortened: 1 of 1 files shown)


thomasmitchell:
lgtm

Issue #656, opened 2019-11-15
[BUG] Targets without jobs can be unhealthy


thomasmitchell:
**Describe the bug**
Targets without jobs can be marked as unhealthy, in which case, they can never become healthy.

**To Reproduce**
1. Make a job that fails
2. Delete the job
3. Look at the data system in the web UI
4. See that it is still red

**Expected behavior**
A target without jobs should either be healthy or... some sort of neutral?

**Screenshots**

**SHIELD versions (please complete the following information):**
 - SHIELD Core: 8.6.0
 - SHIELD CLI: n/a

**Browser version(s) (please complete if reporting a web UI bug):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 70

**Additional context**

Issue #654, opened 2019-11-15
[BUG] v11 migration marks all jobs as failed


thomasmitchell:
**Describe the bug**
All jobs are marked as failing after the migration. Stores probably are too, but those check relatively often. Jobs won't be successful until their next run, which can take awhile. 

**To Reproduce**
1. Deploy 8.5.0
2. Create and run a job successfully
3. Migrate to 8.6.0
4. See it is marked as failing in the UI

**Expected behavior**
It should have maintained its success status.

**Screenshots**

**SHIELD versions (please complete the following information):**
 - SHIELD Core: 8.6.0
 - SHIELD CLI: n/a

**Browser version(s) (please complete if reporting a web UI bug):**
n/a

**Additional context**

thomasmitchell:
Resolved with 8.6.1

PR #657, opened 2019-11-18
Fix database migrations to v11

### LINKED ISSUE #653: [BUG] v11 DB Migration fails when migrating from SHIELD 8.5.0 or earlier
**Describe the bug**
We accidentally deleted the `healthy` member of the jobs table in schema v6. We accidentally added it back for v6 in 8.6.0. If the user has already migrated past this point, they won't get the fix, and the migration to v11 will fail because of the missing `healthy` column.

**To Reproduce**
1. Deploy 8.5.0
2. Migrate to 8.6.0

**Expected behavior**
Should have migrated cleanly.

**Screenshots**

**SHIELD versions (please complete the following information):**
 - SHIELD Core: 8.6.0
 - SHIELD CLI: n/a

**Browser version(s) (please complete if reporting a web UI bug):**
n/a

**Additional context**

sriniketh923:
This pr should fix #653, #654 and #656 . I tested the database migration locally by using the code from 8.5.0 and then upgrading it and it works fine.

### DIFF
--- a/db/schema_v11.go
+++ b/db/schema_v11.go
@@ -7,31 +7,31 @@ func (s v11Schema) Deploy(db *DB) error {
 
 	// set the tenant_uuid column to NOT NULL
 	err = db.Exec(`CREATE TABLE jobs_new (
-	               uuid               UUID PRIMARY KEY,
-	               target_uuid        UUID NOT NULL,
-	               store_uuid         UUID NOT NULL,
-	               tenant_uuid        UUID NOT NULL,
-	               name               TEXT NOT NULL,
-	               summary            TEXT NOT NULL,
-	               schedule           TEXT NOT NULL,
-	               keep_n             INTEGER NOT NULL DEFAULT 0,
-	               keep_days          INTEGER NOT NULL DEFAULT 0,
-	               next_run           INTEGER DEFAULT 0,
-	               priority           INTEGER DEFAULT 50,
-	               paused             BOOLEAN NOT NULL DEFAULT 0,
-	               fixed_key          INTEGER DEFAULT 0,
-	               healthy            BOOLEAN NOT NULL DEFAULT 0
-	             )`)
+			        uuid               UUID PRIMARY KEY,
+					target_uuid        UUID NOT NULL,
+					store_uuid         UUID NOT NULL,
+					tenant_uuid        UUID NOT NULL,
+					name               TEXT NOT NULL,
+					summary            TEXT NOT NULL,
+					schedule           TEXT NOT NULL,
+					keep_n             INTEGER NOT NULL DEFAULT 0,
+					keep_days          INTEGER NOT NULL DEFAULT 0,
+					next_run           INTEGER DEFAULT 0,
+					priority           INTEGER DEFAULT 50,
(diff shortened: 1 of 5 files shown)


db/schema_v11.go
```diff
@@ -7,31 +7,31 @@ func (s v11Schema) Deploy(db *DB) error {
...
-	               keep_n             INTEGER NOT NULL DEFAULT 0,
-	               keep_days          INTEGER NOT NULL DEFAULT 0,
-	               next_run           INTEGER DEFAULT 0,
-	               priority           INTEGER DEFAULT 50,
-	               paused             BOOLEAN NOT NULL DEFAULT 0,
-	               fixed_key          INTEGER DEFAULT 0,
-	               healthy            BOOLEAN NOT NULL DEFAULT 0
-	             )`)
+			        uuid               UUID PRIMARY KEY,
+					target_uuid        UUID NOT NULL,
```
thomasmitchell: This formatting is borked.

thomasmitchell:
Additionally, because the targets being marked as failed with no jobs thing was pre-existing (and there may be existing migrations to schema 11 because it was borked in 5.6.0), we need to make a fixup that patches that up for existing entries.

thomasmitchell:
It also needs to have some handling in 11 either way, because you can't get to 12 if you can't finish 11.

sriniketh923:
I just took out "j.healthy" in v11 and set the default health to 0, which should make the migration from v10 to v11 pass. In v12 all the migrations of the database should be fixed. Even the targets are set to healthy by default and is set to false only when it sees an unhealthy job. So even if a target has no jobs, it should be healthy.

db/schema_v11.go
```diff
@@ -44,6 +44,15 @@ func (s v11Schema) Deploy(db *DB) error {
 		return err
 	}
 
+	jobs, err := db.GetAllJobs(&JobFilter{})
```
thomasmitchell: This is redundant. Put it only in v12.

db/schema_v12.go
```diff
@@ -0,0 +1,115 @@
...
+
+import "fmt"
+
+type v12Schema struct{}
+
+func (s v12Schema) Deploy(db *DB) error {
+	var err error
+
+	// set the tenant_uuid column to NOT NULL
+	err = db.Exec(`CREATE TABLE jobs_new (
```
thomasmitchell: why do we add this table in both 12 and 11?

db/schema_v12.go
```diff
@@ -0,0 +1,76 @@
...
+
+	jobs, err := db.GetAllJobs(&JobFilter{})
+	if err != nil {
+		return nil
+	}
+	for _, job := range jobs {
+		healthy := job.LastTaskStatus == "done"
+		db.UpdateJobHealth(job.UUID, healthy)
+	}
+	tenant := RandomID()
```
jhunt: Why are we setting a random UUID as a default for targets?

sriniketh923: I was running the tests for targets and the tests fail because the UUID was not assigned. I was also looking at the schema_v4 and saw that tennant_uuid is set to a random UUID as a default.

PR #662, opened 2019-12-02
fs plugin - Only print every file name if requested

thomasmitchell:
This was causing a considerable amount of log output which, given our
current architecture, is bringing the SHIELD core to its knees. Not only
does this cause fs jobs to take an extremely long amount of time to
complete (see, 4 days instead of 30 minutes), but it brings the rest of
SHIELD to its knees. This is because all this logging causes a massive
amount of database writes, leading to much lock contention.


thomasmitchell:
This also makes a shared buffer for all the copy operations, such that we don't allocate 32k memory for every file.

PR #674, opened 2020-02-03
Export jrh

sriniketh923:
Import/Export of shield data has been implemented.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@
 /rabbitmq-broker
 /redis-broker
 /s3
-/shieldp
 /swift
 /scality
 /vault
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@
 /rabbitmq-broker
 /redis-broker
 /s3
+/shieldp
 /swift
 /scality
 /vault
--- a/db/export.go
+++ b/db/export.go
@@ -16,10 +16,10 @@ type header struct {
 	N    uint   `json:"n"`
 }
 
-func (db *DB) exportHeader(out *json.Encoder, table string) {
+func (db *DB) exportHeader(out *json.Encoder, table string) error {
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -36,6 +36,7 @@
 /rabbitmq-broker
 /redis-broker
 /s3
+/shieldp
```
jhunt: What is `shieldp`?

db/export.go
```diff
@@ -0,0 +1,521 @@
...
+
+	for r.Next() {
+		v := job{}
+
+		if err = r.Scan(
+			&v.UUID, &v.TargetUUID, &v.StoreUUID, &v.TenantUUID,
+			&v.Name, &v.Summary, &v.Schedule, &v.KeepN, &v.KeepDays,
+			&v.NextRun, &v.Priority, &v.Paused, &v.FixedKey, &v.Healthy); err != nil {
+
+			panic(err)
```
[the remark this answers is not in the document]
jhunt: We might also get away with emiting the error objects into the encoded stream (i.e. `{"error":"..."}`) and either exiting early with some way of signalling to the caller not to continue with more exports, or skipping the problematic thing and letting the caller decide what to do with the data...

In any event, the eventual plugin for backing up shield ought to be able to detect errors in the output stream and change its exit code to `shield-pipe` accordingly.

.gitignore
sriniketh923: This was the shield plugin binary. I'll remove it from this pr.

db/import.go
```diff
@@ -549,6 +563,9 @@ func (db *DB) Import(in *json.Decoder, vault *vault.Client) error {
 				if err != nil {
 					return err
 				}
+			case "errors":
+				err := db.importErrors(in)
+				return err
```
jhunt: You can just return the call, since you're not doing anything with the temporary variable.

PR #679, opened 2020-04-01
Metashield tls

patjones:
Adding two additional but optional fields to allow the meta shield plugin to properly target a shield with tls. 

Previously you could only use the plugin using the agent colocated on the SHIELD box and targetting shieldd on loopback. Now you can specify the SHIELD core CA cert as well as skip ssl validation when configuring a shield backup.


jhunt:
LGTM!

PR #680, opened 2020-04-10
Bootstrap Restore

patjones:
The goal of this PR is to update the SHIELD first-time login restore functionality. Users will now be able to properly restore a SHIELD using an archive created using the new metashield plugin.

This required fundamental changes to the process of recovering SHIELD now that the backup/restore no longer uses the fs plugin. Some of these changes include having to initialize shield prior to restoring with a new master password and fixed key, no longer needing to restart shield after a restore and more.


thomasmitchell:
is this good to merge? @jhunt

jhunt:
go for it

PR #695, opened 2020-07-02
fix ordering of vault status checks even more

thomasmitchell:
This would behave poorly if shieldd had been restarted but not Vault - the Vault would not be uninitialized or sealed, so the health check wouldn't return an error, and so the function would return Ready. But we actually wouldn't have a token yet, so backups would fail.

I had screwed up the ordering when I tried to fix the previous ordering, which would check for token validity before checking for vault seal/uninit status.

To validate that this works, it needs to be tested that the unseals and backs up as expected
1. on a clean install
2. after a restart of both vault and shield
3. after a restart of just shield
4. after a restart of just vault

### DIFF
--- a/core/vault/client.go
+++ b/core/vault/client.go
@@ -66,20 +66,27 @@ const (
 
 func (c *Client) Status() (Status, error) {
 	err := c.vault.Health(false)
-	if err == nil {
-		return Ready, nil
-	}
 	if vaultkv.IsUninitialized(err) {
 		return Blank, nil
 	}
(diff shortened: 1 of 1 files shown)


sriniketh923:
Finished testing this and good to me merged!

jhunt:
Is the desire to merge this about short-term support for 8.x?  Vault integration goes away in v9...

sriniketh923:
Yes, it's for 8.x.

PR #707, opened 2021-11-30
Account for M1 arm64 within api tests. ${PWD} NOT $(pwd). [[ ]] NOT [ ].

wayneeseguin:


wayneeseguin:
Missed a spot, will fix and update PR.

xiujiao:
LGTM

PR #710, opened 2022-02-02
Adding Mac ARM64 functionality to shield

antitux:
This adds functionality to shield to build native arm64 binaries and updates the shield UI to point to both intel and arm64 binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ DOCKER_TAG   ?= dev
 EDGE         ?= edge
 
 # Everything; this is the default behavior
+LDFLAGS := -X main.Version=$(VERSION)
 all: build test
 build: format shieldd shield shield-agent shield-schema shield-crypt shield-report plugins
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -125,7 +127,8 @@ release:
 	cd artifacts && for x in shield-server-*; do \
 	  cp -a ../web/htdocs $$x/webui; \
 	  mkdir -p $$x/webui/cli/linux; cp ../artifacts/shield-linux-amd64   $$x/webui/cli/linux/shield; \
-	  mkdir -p $$x/webui/cli/mac;   cp ../artifacts/shield-darwin-amd64  $$x/webui/cli/mac/shield; \
+	  mkdir -p $$x/webui/cli/mac-amd64;   cp ../artifacts/shield-darwin-amd64  $$x/webui/cli/mac-amd64/shield; \
+	  mkdir -p $$x/webui/cli/mac-arm64; cp ../artifacts/shield-darwin-arm64 $$x/webui/cli/mac-arm64/shield; \
```
wayneeseguin, line 131: Perhaps it would be cleaner to do mac/{amd64,arm64} as subdirectories?

wayneeseguin: Additionally I think we should stick to uname output and have it be `darwin/$ARCH`

wayneeseguin: More clear statement: it should be `darwin/{amd64/arm64}` not `mac-*`

antitux: I'm fine with that too.

PR #711, opened 2022-02-02
Added backup job retry feature

pururval:


cmd/shield/help.go
```diff
@@ -336,6 +336,9 @@ func ShowHelp(command string) {
 		fmt.Printf("                  in days (7d) or weeks (5w).\n")
 		fmt.Printf("                  This field is @W{required}.\n")
 		fmt.Printf("\n")
+		fmt.Printf("  --retries       How many times to retry the backup job if it fails. Can be given\n")
+		fmt.Printf("                  in times (7).\n")
```
wayneeseguin, line 340: The default should be in parens, is it 7 or is it 3 as the code below?

cmd/shield/help.go
```diff
@@ -1779,6 +1783,9 @@ func ShowHelp(command string) {
 		fmt.Printf("  --retain        How long to keep backup archives.  Can be given\n")
 		fmt.Printf("                  in days (7d) or weeks (5w).\n")
 		fmt.Printf("\n")
+		fmt.Printf("  --retries       How many times to retry the backup job if it fails. Can be given\n")
+		fmt.Printf("                  in times (7).\n")
```
wayneeseguin, line 1787: Same as above 7 vs 3 for default.

cmd/shield/help/create-job
```diff
@@ -30,6 +30,9 @@ USAGE: @G{shield} create-job --tenant @Y{TENANT} [OPTIONS]
                   in days (7d) or weeks (5w).
                   This field is @W{required}.
 
+  --retries       How many times to retry the backup job if it fails.  Can be given
+                  in times (7d).
```
wayneeseguin, line 34: Same comment

cmd/shield/help/create-job
```diff
@@ -50,7 +53,8 @@ USAGE: @G{shield} create-job --tenant @Y{TENANT} [OPTIONS]
       @Y{--target}    "Important Files" \
       @Y{--store}     "Amazon S3"       \
       @Y{--schedule}  "daily 4am"       \
-      @Y{--retain}    14d
+      @Y{--retain}    14d               \
+      @Y{--retries}    2
```
wayneeseguin, line 57: This one has 2 vs 3 or 7, we should be consistent.

cmd/shield/help/update-job
```diff
@@ -26,6 +26,9 @@ USAGE: @G{shield} update-job --tenant @Y{TENANT} [OPTIONS] @Y{NAME-OR-UUID}
   --retain        How long to keep backup archives.  Can be given
                   in days (7d) or weeks (5w).
 
+  --retries       How many times to retry the backup job if it fails.  Can be given
+                  in times (7d).
```
wayneeseguin, line 30: Same comment.

cmd/shield/import.go
```diff
@@ -160,6 +162,9 @@ func (m *ImportManifest) Normalize() error {
 				if job.Retain == "" {
 					return fmt.Errorf("Tenant '%s', data system '%s', job '%s' is missing its `retain' attribute", tenant.Name, system.Name, job.Name)
 				}
+				if job.Retries == 0 {
+					return fmt.Errorf("Tenant '%s', data system '%s', job '%s' is missing its `retries' attribute", tenant.Name, system.Name, job.Name)
```
wayneeseguin, line 166: Will this cause a problem for someone who upgraded with active jobs in play that will default to `0` for retries?

cmd/shield/main.go
```diff
@@ -802,12 +804,14 @@ tenants:
...
             storage:  Local Storage
             retain:   4d
+            retries:  3
 
           - name:     Weekly
             when:     sundays at 2:45am
             paused:   yes
             storage:  Local Storage
             retain:   28d
+            retries:  2
```
wayneeseguin, line 814: What's the logic for 2 vs 3 on these?

core/scheduler/chore.go
```diff
@@ -116,8 +116,26 @@ func (w *Worker) Execute(chore Chore) {
...
 	go func() {
 		chore.Do(chore)
+
+		job, err := w.db.GetJob(task.JobUUID)
+		if err != nil {
+			panic(fmt.Errorf("failed to retrieve job '%s' from database: %s", task.JobUUID, err))
+		}
+		retries := job.Retries
+		log.Errorf("Retries: %d", chore)
+		if rc != 0 && retries > 0 {
```
wayneeseguin, line 132: Are we making sure to store the retries # in the database in case the process exits prematurely? Eg. we should build failure handling into the retries. With this in mind perhaps it would be a better approach to modify the task running selection to account for # retries < retries. It's OK to do this as a feature enhancement as well as long as it is accounted for.

db/bus_test.go
```diff
@@ -201,6 +203,7 @@ var _ = Describe("MessageBus Database Integration", func() {
 					Paused:   true,
 					FixedKey: false,
 					Healthy:  false,
+					Retries: 4,
```
wayneeseguin, line 206: Same question, why some are 3 vs 4 vs 2.

wayneeseguin (changes requested):
Overall looks good on the right path. See comments about consistency and feature approach.

cmd/shield/help.go
pururval: I'll change this to 3.

cmd/shield/help.go
pururval: I'll change this to 3.

cmd/shield/help/create-job
pururval: I'll change this to 3.

cmd/shield/help/create-job
pururval: I'll change this to 3.

cmd/shield/help/update-job
pururval: I'll change this to 3.

cmd/shield/main.go
pururval: I'll change this to 3.

db/bus_test.go
pururval: I'll change this to 3. Some of them are docs or local testing files, none of those are really used.

core/scheduler/chore.go
pururval: The retries configured for the job by the user are stored in the database, but it does make sense to look into the "current retry". I'm thinking, retries itself is a failure-handle and looking at the current lightweight nature of SHIELD, would it be a far stretch? Also, it will increase the complexity of the state of tasks which again would need modifications in other operations- unfailed vs half-failed vs retry failed. Another thing which I think would be more important is scheduling retries at different time intervals. Generally if something fails now, it usually is going to fail in the next few minutes (saying after looking at the enormous number of fail-checks in SHIELD code). That leaves network or storage issue to be the major factor which would not be different in same moment.  Maybe having each retry happen after a certain time interval would make it more useful?

pururval: Created a [JIRA ticket](https://starkandwayne.atlassian.net/browse/SHIELD-6)

PR #713, opened 2022-02-10
Adding support for old Schema getAllJobs call

pururval:
Schema6 and Schema 12 use getAllJobs which make adding new variables in jobs struct difficult. Adding a separate getAllJobsV6() to support old schema as well has new schemaV13 with configurable variable "retries" for job retry on fail.


wayneeseguin (commented):
If the tests pass LGTM.

PR #716, opened 2022-06-14
Renamed RabbitMQ Plugin and added / trim

pururval:
Renamed the RabbitMQ Broker Backup Plugin to remove 'Pivotal' from it. And added a forward slash truncate for the RabbitMQ dashboard url  user input when applicable.


norman-abramovitz (approved):
LGTM

PR #722, opened 2022-08-25
Added releases notes for the bug fix- Login screen in Github OAuth

pururval:
Bug fix for the login screen of Shield not showing Github OAuth button.


norman-abramovitz (approved):
The actual fix was done back on Aug 10th and we needed the release notes to finish things off.

PR #725, opened 2022-10-04
Adding okta integration.

pururval:
Adding an integration in SHIELD to support authentication with OKTA. Added a file auth_provider_okta.go for majority of okta authentication functions and also a documentation at docs/ops/okta-auth.md for information on setting up the requirements for the integration. This is a base implementation that works on a dev-okta account and may need further tweaking depending on client environments and other variables.


core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	return ll
+}
+
+func (p *OktaAuthProvider) Initiate(r *route.Request) {
+
+	var state = randstr.Hex(16)
+
+	var redirectPath string
+
+	q := r.Req.URL.Query()
```
christian-roggia, line 116: Is there a reason why we us the request `URL.Query` instead of a new instance of Values? https://pkg.go.dev/net/url#Values

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	q.Add("client_id", p.ClientID)
+	q.Add("response_type", "code")
+	q.Add("response_mode", "query")
+	q.Add("scope", "openid profile email groups")
+	q.Add("redirect_uri", p.DeploymentURI + "/auth/okta/redir") //ex: https://shield.starkandwayne.com/auth/okta/redir
+	q.Add("state", state)
+	// q.Add("nonce", nonce)
+
+	redirectPath = p.OktaDomain + "/oauth2/" + p.AuthorizationServer + "/v1/authorize?" + q.Encode()
```
christian-roggia, line 125: For completeness and cleanliness, whenever dealing with URLs, consider using the URL{} struct from the url standard package. https://pkg.go.dev/net/url#URL

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+func (p *OktaAuthProvider) HandleRedirect(r *route.Request) *db.User {
+	
+	// The login is achieved through the authorization code flow, where the user 
+	// is redirected to the Okta-Hosted login page. After the user authenticates 
+	// they are redirected back to the application with an access code that is 
+	// then exchanged for an access_token.
+	
+	// log.Debugf("DEBUG::r *route.Request is [GET /auth/okta/redir]")
+
+	q := r.Req.URL.Query()
```
christian-roggia, line 140: Same here as before.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	q.Add("grant_type", "authorization_code")
+	q.Set("code", r.Param("code", ""))
+	q.Add("redirect_uri", p.DeploymentURI + "/auth/okta/redir") //ex: https://shield.starkandwayne.com/auth/okta/redir
+	q.Add("response_type", "token")
+
+	url := p.OktaDomain + "/oauth2/" + p.AuthorizationServer + "/v1/token?" + q.Encode()
+
+	authHeader := base64.StdEncoding.EncodeToString([]byte(p.ClientID + ":" + p.ClientSecret))
```
christian-roggia, line 148: Use https://pkg.go.dev/net/http#Request.SetBasicAuth to set the `Authorization` header instead.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+
+	req, _ := http.NewRequest("POST", url, bytes.NewReader([]byte("")))
+	h := req.Header
+	h.Add("Authorization", "Basic "+authHeader)
+	h.Add("Accept", "application/json")
+	h.Add("Content-Type", "application/x-www-form-urlencoded")
+	h.Add("Connection", "close")
```
christian-roggia, line 155: I am fairly confident that `Connection` and `Content-Length` headers are ignored and automatically set by the http client, I could be wrong though.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	h.Add("Content-Length", "0")
+
+	client := &http.Client{}
+	resp, _ := client.Do(req)
+	body, _ := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
+	var exchange Exchange
+	json.Unmarshal(body, &exchange)
```
christian-roggia, line 163: Do not silence errors.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	req, _ := http.NewRequest("POST", url, bytes.NewReader([]byte("")))
+	h := req.Header
+	h.Add("Authorization", "Basic "+authHeader)
+	h.Add("Accept", "application/json")
+	h.Add("Content-Type", "application/x-www-form-urlencoded")
+	h.Add("Connection", "close")
+	h.Add("Content-Length", "0")
+
+	client := &http.Client{}
+	resp, _ := client.Do(req)
```
christian-roggia, line 159: Do not silence errors.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	body, _ := ioutil.ReadAll(resp.Body)
```
christian-roggia, line 160: Do not silence errors.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil
+	}
+
+	// Now we succussfully exhchanged the auth code for an openID token which is stored in our struct "exchange"
+	// Next we verify the token(s) and extract our okta claims from it (i.e. groups)
+	// Goal is to get {name, account, and orgs} from this token to do the mapping. 
+
+	id_token, verificationError := p.verifyToken(exchange.IdToken, "id")
+	if verificationError != nil {
+		fmt.Println(verificationError)
```
christian-roggia, line 176: Shouldn't we return here as verification failed?

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	access_token, verificationError := p.verifyToken(exchange.AccessToken, "access")
+	if verificationError != nil {
+		fmt.Println(verificationError)
```
christian-roggia, line 180: Shouldn't we return here as verification failed?

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	for _, org_name := range org_names {
+		orgs[org_name] = make([]string, 0)
+	}														//ex: orgs- [map[okta:[]]] - [map[string][]string]
+
+	g, err := json.Marshal(groups_claim) 					//ex: g-[["Everyone","User","Admin"]] - [[]uint8]
+	if err != nil {
+		fmt.Println(err)
+	}
+	groups_list :=  make([]string, 0)
+	json.Unmarshal(g, &groups_list)							//ex: groups_list-[[Everyone User Admin]] - [[]string]
```
christian-roggia, line 204: Do not silence errors.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+
+	if user == nil {
+		user = &db.User{
+			UUID:    uuid.NewRandom().String(),
+			Name:    name,
+			Account: account,
+			Backend: p.Identifier,
+			SysRole: "",
+		}
+		p.core.db.CreateUser(user)
```
christian-roggia, line 243: Do not silence errors.

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+
+							return nil
+						}
+						continue Mapping
+					}
+
+					for _, group := range groups {
+						if match.Group == group {
+							if !p.Assign(user, candidate.Tenant, match.Role) {
+
```
christian-roggia, line 265: Shouldn't we log an error here?

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+							}
+							continue Mapping
+						}
+					}
+				}
+			}
+		}
+	}
+	if !p.SaveAssignments(p.core.db, user) {
+		return nil
```
christian-roggia, line 276: Shouldn't we log an error here?

core/auth_provider_okta.go
```diff
@@ -0,0 +1,311 @@
...
+	p.ClearAssignments()
+	
+Mapping:
+	for _, candidate := range p.Mapping {
+		for org, groups := range orgs {
+			if candidate.Okta == org {
+				for _, match := range candidate.Rights {
+					if match.Group == "" {
+						if !p.Assign(user, candidate.Tenant, match.Role) {
+
```
christian-roggia, line 256: Shouldn't we log an error here?

christian-roggia (changes requested):

core/auth_provider_okta.go
pururval: Empty group would be the default case, in which user is given the base role (operator).

core/auth_provider_okta.go
pururval: Similar pattern to https://github.com/shieldproject/shield/blob/0509411d3ae64b823db8d119ab19a34dfceaee2b/core/auth_provider_github.go#L185

core/auth_provider_okta.go
christian-roggia: Then shouldn't be the user returned instead of `nil`?

core/auth_provider_okta.go
pururval: SetBasicAuth uses `r.Header.Set("Authorization", "Basic "+basicAuth(username, password))` but we have `base64.StdEncoding.EncodeToString([]byte(p.ClientID + ":" + p.ClientSecret))`

core/auth_provider_okta.go
pururval: Done

core/auth_provider_okta.go
pururval: Done

core/auth_provider_okta.go
pururval: Done

core/auth_provider_okta.go
pururval: Done

core/auth_provider_okta.go
pururval: Done

core/auth_provider_okta.go
pururval: Similar pattern to https://github.com/shieldproject/shield/blob/0509411d3ae64b823db8d119ab19a34dfceaee2b/core/auth_provider_github.go#L185

core/auth_provider_okta.go
pururval: url.Values{} still needs to add each parameter in a single line. I can maybe make a custom struct for it?

core/auth_provider_okta.go
pururval: type URL struct {
	Scheme      [string](https://pkg.go.dev/builtin#string)
	Opaque      [string](https://pkg.go.dev/builtin#string)    // encoded opaque data
	User        *[Userinfo](https://pkg.go.dev/net/url#Userinfo) // username and password information
	Host        [string](https://pkg.go.dev/builtin#string)    // host or host:port
	Path        [string](https://pkg.go.dev/builtin#string)    // path (relative paths may omit leading slash)
	RawPath     [string](https://pkg.go.dev/builtin#string)    // encoded path hint (see EscapedPath method)
	OmitHost    [bool](https://pkg.go.dev/builtin#bool)      // do not emit empty host (authority)
	ForceQuery  [bool](https://pkg.go.dev/builtin#bool)      // append a query ('?') even if RawQuery is empty
	RawQuery    [string](https://pkg.go.dev/builtin#string)    // encoded query values, without '?'
	Fragment    [string](https://pkg.go.dev/builtin#string)    // fragment for references, without '#'
	RawFragment [string](https://pkg.go.dev/builtin#string)    // encoded fragment hint (see EscapedFragment method)
}

It does not coincide with the values we have.

core/auth_provider_okta.go
christian-roggia: You can simply do like the following:
```golang
...
	q := url.Values{}
	q.Add("client_id", p.ClientID)
	q.Add("response_type", "code")
...
```

core/auth_provider_okta.go
christian-roggia: It would be like the following:
```golang
u := url.URL{
  Scheme: strings.Split(p.OktaDomain, "://")[0],
  Host: strings.Split(p.OktaDomain, "://")[1],
  Path: url.JoinPath("/oauth2/", p.AuthorizationServer, "/v1/authorize"),
  RawQuery: q.Encode(),
}
```

Maybe consider splitting scheme and host in the `OktaAuthProvider` definition.

core/auth_provider_okta.go
christian-roggia: Then we should keep it as it is, nice catch.

christian-roggia:
LGTM

wayneeseguin:
Merge it

christian-roggia (approved):

PR #743, opened 2023-07-21
Consul 1 14 5

itsouvalas:
Updates consul and its dependencies to 1.14.5

Breaking changes:

Removes Azure addon is disabled due to a rewrite on Azure's go module. Future releases will rewrite and re-instate the addon.


dennisjbell:
@itsouvalas please update the vendored libraries

Issue #744, opened 2023-07-25
Need to be able to trigger cloud storage health check after rectifying the issue


dennisjbell:
Currently no (obvious) way to tell cloud storage to re-check its health when its failing.

pururval:
@dennisjbell Please let me know if PR #748  works or if you have any suggestions for improvements. Thank you!

PR #748, opened 2023-12-06
Added a health check button for Stores.

pururval:
As a fix for [issue-744](https://github.com/shieldproject/shield/issues/744), add a button `Test` on stores page which basically triggers a health check task for that store.

### DIFF
--- a/core/api_v2.go
+++ b/core/api_v2.go
@@ -2610,6 +2610,33 @@ func (c *Core) v2API() *route.Router {
 
 		r.OK(store)
 	})
+	r.Dispatch("POST /v2/tenants/:uuid/stores/:uuid/test", func(r *route.Request) { // {{{
+		// if c.IsNotTenantOperator(r, r.Args[1]) {
+		// 	return
+		// }
+
+		store, err := c.db.GetStore(r.Args[2])
+		if err != nil {
+			r.Fail(route.Oops(err, "Unable to retrieve storage system information"))
+			return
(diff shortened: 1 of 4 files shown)


wayneeseguin:
Thanks for this, @pururval — and apologies for how long it sat.

This has landed on `develop` as of 46a74929, with your original commit
preserved as-authored. It went in via #815 rather than by merging this PR
directly, because a few things needed fixing on top and your fork's branch had
gone quiet. The follow-ups are separate commits, so your work stays intact in
the history.

What changed on top of your commit:

- The authorization check in the new route was commented out, which let any
  authenticated session queue a health check against another tenant's store.
  Enabled it, requiring the tenant engineer role to match the sibling store
  routes.

- A failed `CreateTestStoreTask` was logged as non-critical and still returned
  success. That's right on store *create*, where the test is incidental — but
  here scheduling is the whole point, so it now fails the request.

- The store view offers the button to system engineers too, so
  `POST /v2/global/stores/:uuid/test` was added along with the missing
  `#!/admin/stores/test` dispatch case — without it the button did nothing in
  the admin context.

- The success handler navigated to `#!/stores/store:uuid:` with the uuid
  omitted, landing on a broken store view.

- The modal's `safe`/`danger` button classes were inverted relative to the
  other confirm modals, colouring Cancel as the destructive choice.

Closing in favour of #815. Thanks again for the contribution — this closes out
#744.

PR #757, opened 2024-07-15
Update Go version and dependancy. Use Quay for docker images

krutten:
Update Go version
Update Docker Repository


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   vault:
-    image: vault:1.0.2
+    image: hashicorp/vault:1.14
```
norman-abramovitz, line 4: Why version 1.14 vs 1.17?

norman-abramovitz (commented):
I ignored the vendor changes.  What is the reason to go with 1.21 instead of 1.22?

docker-compose.yml
krutten: GoLang just released 1.22.x and 1.21.12 is still in it's support cycle and is N-1

Vault has breaking changes in newer versions and we will be updating the Demo but more testing is needed.  The docker-compose.yml is not used in production deployments

dennisjbell (approved):

PR #775, opened 2025-01-23
[WIP] Fix the initialization flow for newer Vault versions

krutten:
Newer version of Vault do not create the secret mount by default causing a race condition on first creation.  This ensures the mount is created during setup.


norman-abramovitz (commented):
The release notes should make it clear that Shield will only work on Vault KV Version=2 now and into the future.

norman-abramovitz (approved):
LGTM

PR #797, opened 2025-09-25
Update S3 plugin to use IMDSv2

krutten:
Updates the S3 plugin to support IMDSv2 when "Use Instance Profile" enabled


wayneeseguin (approved):
Tested this in an env looking good

PR #813, opened 2026-07-27
Close SSH-stack and stdlib CVEs (x/crypto v0.52.0, Go 1.25.12)

itsouvalas:
A security patch over v9.0.1, closing the reachable SSH-stack and
standard-library advisories and adding a probe that guards the SSH
channel against future dependency drift.

**golang.org/x/crypto v0.50.0 → v0.52.0.** SHIELD imports
golang.org/x/crypto/ssh first-party in both roles — the shield-agent is
an SSH server, core is the client — so these are reachable. Clears eight
advisories: CVE-2026-39828, -39829, -39830, -39831, -39834, -39835,
-46595, -46597.

**Go build floor 1.25.5 → 1.25.12.** Closes three stdlib DoS advisories:
GO-2026-4977 and -4986 (net/mail) and -4342 (archive/zip).

**SSH-channel probe (`agent/xcrypto_probe_test.go`).** Stands up the real
shield-agent and drives a task through core's own client path, asserting
the handshake and round-trip and pinning the negotiated crypto — nothing
else exercises this channel end to end with the production client config.


wayneeseguin (approved):
Approving after independent verification. Analysis below.

## Vulnerability claims — verified by govulncheck

Ran \`govulncheck -mode=source ./...\` against both \`develop\` (29c6d3f) and this branch (6d1ca6a), comparing only findings with a real call trace:

| develop (called) | this branch |
|---|---|
| GO-2026-5013 / CVE-2026-46597 | gone |
| GO-2026-5014 / CVE-2026-39828 | gone |
| GO-2026-5015 / CVE-2026-39835 | gone |
| GO-2026-5017 / CVE-2026-39830 | gone |
| GO-2026-5018 / CVE-2026-39829 | gone |
| GO-2026-5019 / CVE-2026-39831 | gone |
| GO-2026-5020 / CVE-2026-39834 | gone |
| GO-2026-5023 / CVE-2026-46595 | gone |

All eight list `fixed: 0.52.0` in the Go vulnerability database — the bump is the minimum that clears them, not an over-bump. The eight CVE IDs in the PR body match exactly.

Reachability confirmed in both roles, as claimed. govulncheck traces on develop:

```
agent/agent.go:76:45:        agent.Agent.ServeOne calls ssh.NewServerConn
core/fabric/legacy.go:146:25: fabric.Execute calls ssh.Dial
```

Four `github.com/docker/docker` findings remain on both branches (GO-2026-4883, -4887, -5617, -5668). All are `Fixed in: N/A` upstream — correctly out of scope here.

Stdlib floor checks out too. Fix versions from the vuln DB: GO-2026-4977 and GO-2026-4986 (net/mail) at 1.25.10, GO-2026-4342 (archive/zip) at 1.25.6. A floor of 1.25.12 clears all three. 1.25.12 is the current latest 1.25 patch, so `golang:1.25-bookworm` in the Dockerfile already resolves to it — no toolchain download is introduced into the image build.

## Build and vendor integrity

- `go mod verify` — all modules verified
- `go build -mod=vendor ./...` — clean
- `go mod vendor` regenerated — **zero diff**, so the vendored tree matches go.mod/go.sum exactly
- go.mod/go.sum hashes are internally consistent; x/net, x/sys, x/term, x/text move only as far as x/crypto v0.52.0 requires

## Tests

New probes pass, all six:

```
go test -mod=vendor ./agent/ -run TestXCryptoProbe   →  6 passed
```

No regression. `TestAgent` ("Command Runner works", exit 127) fails identically on `develop` and on this branch — pre-existing PATH fault, untouched by this PR. Other packages (db, plugin, timespec) pass on both.

`go vet` reports only pre-existing findings in files this PR does not touch (plugin/fs/plugin.go, db/tenant_test.go).

## Code review

The only production change is 8/-1 in `core/core.go`: the default MAC list is lifted into an exported `DefaultLegacyAgentMACs` so the probe reads core own list rather than a copy. Behaviour is identical. The read-only caveat in the doc comment is the right call — the slice does back a config field, so an in-place append would be a live-data bug.

The probe itself is unusually careful. Worth calling out specifically:

- Probe 4 is a genuine negative control. Under the default AEAD cipher the MAC pin is inert, so a happy-path round trip proves nothing about MAC negotiation; forcing `aes128-ctr` makes it load-bearing and the assertion meaningful. Few probes bother to prove they can see the failure they claim to guard.

- Probe 5 exercises the exact `CertChecker` shape CVE-2026-39835 describes, in-process, so a handshake panic takes the test down with it. That is a real assertion, not a proxy for one.

- The unbounded serve loop and the `SHIELD_AGENT_LISTEN_ADDRESS` override are both correct, and the reasoning in the comments matches what the code does.

## One maintenance note, not a blocker

`probeFingerprintBaseline` pins the negotiated KEX, host key, cipher, MAC, and host-key fingerprint, and the gate is on by default. That is deliberate and I agree with defaulting it on — an opt-in gate would not fire under `make go-tests`.

The cost is that any future x/crypto bump that shifts algorithm preference fails this test until someone re-measures, and the failure will land on whoever runs that bump rather than on whoever caused it. The failure message does point at the fix, which mitigates it. Flagging so it is a known cost rather than a surprise.

Also noting `concourse-ci/status` has been pending since the branch was pushed and has not moved. `ci/pipeline/jobs/test-pr.yml.disabled` is disabled, so PRs are not tested by the pipeline; branch protection does not require status checks. Nothing verified this change in CI — the evidence above is from a local run.

